PR #270, opened 2014-12-11
Fix: Removed unused methods inside ViewHelpers

ins0:
This removes unused methods inside the viewhelper


localheinz:
@Ocramius 

Shall I have a look?

Ocramius:
@localheinz ok, assigning it to you, but please check that the prod site is running first.

localheinz:
Will do, @Ocramius !

localheinz:
@GeeH 

I think the production site is running ok, you just mentioned it a second ago!

PR #279, opened 2014-12-13
Fix: removed rewrite base

ins0:
Removed rewrite base - not necessary and makes it easier for contributors to run this repo locally without forced to setup a virtual host.


localheinz:
@ins0 

Needs a rebase, unfortunately!

ins0:
@localheinz rebased

localheinz:
Thank you, @ins0!

PR #289, opened 2014-12-16
github registration listener refactoring + tests

lku:
As promised in #286 this PR moves registration listener from closure in `User\Module` to separate listener class. There is both unit and integration test case to assure that listener is attached and `registerViaProvider` event triggered in `ScnSocialAuth` module.

I still have to find a way how to update existing local user profile from actual GitHub data after login.

@Ocramius could you please provide code review?

### DIFF
--- a/module/User/config/module.config.php
+++ b/module/User/config/module.config.php
@@ -24,4 +24,9 @@ return [
             'userOrganizations' => 'User\View\Helper\UserOrganizationsFactory',
         ],
     ],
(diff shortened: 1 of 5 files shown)


localheinz:
@lku 

Needs a rebase, can you have a look?

lku:
@localheinz 
Done. There was just a little conflict in User module config (because of switch to short array syntax).

localheinz:
Thank you, @lku!

PR #294, opened 2014-12-22
Enhancement: Leverage class keyword, remove PHP5.4 from build matrix

localheinz:
Because when we've got PHP5.5, we should use it.

/cc @ins0


localheinz:
Closing this one, it's doing too much.

PR #298, opened 2015-01-11
Fix: Remove duplicate entry in array of modules

localheinz:
The `Application` module is registered a few lines further down the line again, once should be enough.

### DIFF
--- a/config/application.config.php
+++ b/config/application.config.php
@@ -3,7 +3,6 @@ ini_set('display_errors', 1);
 
 return array(
     'modules' => array(
-        'Application',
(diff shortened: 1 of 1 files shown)


Ocramius:
Is it supposed to not be the first module?

ins0:
neither - don't see any reason it should appear at the beginning as well as in the middle. so this change looks fine to me

Ocramius:
Thanks @localheinz!

localheinz:
Thanks for merging so quickly, @Ocramius!

PR #299, opened 2015-01-11
Fix: Removed the cache logic from module controller

ins0:
the github client already have a cache logic implement and this stuff can dumped out of the controller logic


Ocramius:
All removals: I like!

localheinz:
Thank you, @ins0!

@Ocramius We've discussed this before, so I'll merge it as soon as we've got a green light.

PR #303, opened 2015-01-11
Fix: composer.lock never seems up to date

localheinz:
As it says in the title.


GeeH:
Do we still need this @localheinz

PR #317, opened 2015-01-23
Fix: added github repo exists check

### LINKED ISSUE #158: Some modules throw exceptions on modules.zendframework.com?
I'm not sure if it's just a temporary problem or related to this specific module:

<http://modules.zendframework.com/weierophinney/PhlySimplePage>

@weierophinney  @EvanDotPro

zf2 modules - 2013-11-06_20 51 30

### LINKED ISSUE #182: Error page on access this link:-  http://modules.zendframework.com/MohitOssCube/ZF2Booster
I have changed my user name in git hub after that, when I am access the following URL 

http://modules.zendframework.com/MohitOssCube/ZF2Booster

an error exception comes please verify the same. 
screenshot from 2014-05-22 15 29 24

ins0:
fixes #182 
fixes #158 

but we need a solution in the future where such repos gets deleted from the db

### DIFF
--- a/module/ZfModule/src/ZfModule/Controller/IndexController.php
+++ b/module/ZfModule/src/ZfModule/Controller/IndexController.php
@@ -49,12 +49,13 @@ class IndexController extends AbstractActionController
         $module = $this->params()->fromRoute('module', null);
 
         $result = $this->moduleMapper->findByName($module);
-        if (!$result) {
+        $repository = $this->repositoryRetriever->getUserRepositoryMetadata($vendor, $module);
+
+        if (!$result || !$repository) {
             $this->getResponse()->setStatusCode(Http\Response::STATUS_CODE_404);
             return;
         }
 
-        $repository = $this->repositoryRetriever->getUserRepositoryMetadata($vendor, $module);
         $readme = $this->repositoryRetriever->getRepositoryFileContent($vendor, $module, 'README.md');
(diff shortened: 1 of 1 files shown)


module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -49,12 +49,13 @@ public function viewAction()
         $module = $this->params()->fromRoute('module', null);
 
         $result = $this->moduleMapper->findByName($module);
-        if (!$result) {
+        $repository = $this->repositoryRetriever->getUserRepositoryMetadata($vendor, $module);
+
+        if (!$result || !$repository) {
             $this->getResponse()->setStatusCode(Http\Response::STATUS_CODE_404);
             return;
```
GeeH: You should return the response here to short circuit the rest of the dispatch process.

ins0: yeah will recode the logic so in case a repository is not exists we can save 1 api call

GeeH:
I think it's absolutely appropriate that our site 404s if the github repo doesn't exists. In an ideal world we would remove the module from our DB too (or flag it as hidden).

ins0:
the only problem i see what if the github api is down this would leed to a repository false flag - maybe some crontab that checks flagged repos one a week and delete them from the db

PR #318, opened 2015-01-23
Fix phpdocs

snapshotpl:


ins0:
the current comments are inline ``/*`` but the suggested would be docblock `/**``  comments. so the current comment should be fine. no?

localheinz:
PHPStorm's intention adds `/**`, but `/*` is sufficient for inline comments, I believe.

Ocramius:
Note that `/**` is a docblock comment, not a normal comment. Don't use it except for docblocks

GeeH:
The comments are correct as they are `/*` is right.

PR #319, opened 2015-01-23
Remove unused and duplicate variable

snapshotpl:


Ocramius:
@snapshotpl thanks!

module/Application/src/Application/Module.php
```diff
@@ -20,13 +20,11 @@ class Module
 {
     public function onBootstrap($e)
     {
-        $e->getApplication()->getServiceManager()->get('translator');
         $eventManager        = $e->getApplication()->getEventManager();
         $moduleRouteListener = new ModuleRouteListener();
```
snapshotpl: Question to @Ocramius . Are we need this listener? Its attached twice: here and ZfModule module.php

PR #321, opened 2015-01-23
Remove composer minimum stability

snapshotpl:
With composer minimum stabilty dev:
* zf and many others is loaded from master (unstable),
* loading time from branch (no tag) it's sooooo long...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,6 @@
             "UserTest": "module/User/test/",
             "ZfModuleTest": "module/ZfModule/test/"
         }
-    },
-    "minimum-stability": "dev"
+    }
 }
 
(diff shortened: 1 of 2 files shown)


localheinz:
:+1: for also actually pinning to concrete versions, unless necessary to pin to a branch.

GeeH:
Have you tested this? I would expect a corrisponding `composer.lock` update (unless this doesn't change loaded versions).

snapshotpl:
Everything working fine: manual and unit tests. I see a huge different in speed when I run `composer install` becouse with stable option composer download from tag (zip), without stable download from branch.

GeeH:
:+1: 

Thanks for your time :)

PR #322, opened 2015-01-23
Improve views

snapshotpl:


module/Application/view/application/index/index.phtml
```diff
@@ -1,24 +1,20 @@
 <?php $this->headTitle(sprintf('Browse %d modules for Zend Framework 2', $this->totalModules())); ?>
+<?php $this->inlineScript()->appendScript('$(\'#live-search input[name="query"]\').liveSearch({url: "/live-search?query="});'); ?>
```
GeeH: I believe inline scripts should go at the bottom of the page.

GeeH:
I had a quick look and I don't see the differences in what this is changing, can you clarify please?

GeeH:
I get it now sorry, it's semantic changes which are welcome, if you can take a look at the inline comment above I will merge :)

module/Application/view/application/index/index.phtml
snapshotpl: Why? This is only append to container, not print.

ins0: i prefer to write this things on the top of view templates too

GeeH: Not relevant, the inlineScript should be at the bottom of the page so that it doesn't block page rendering on some browsers. It's a best practice.

ins0: the code will be inserted in layout with ``$this->inlineScript`` not in the template file itself

GeeH: Roger, merging, thanks sorry for confusion I have hang over.

PR #324, opened 2015-01-24
Move factories to separate file, format fixes

snapshotpl:


GeeH:
Epic. Thanks.

gianarb:
Maybe we can override it with this key 
https://developer.github.com/v3/oauth/#redirect-urls

PR #328, opened 2015-01-24
Update Readme.md

snapshotpl:


localheinz:
Thanks, @snapshotpl!

I think the whole thing could use an update anyway.

PR #332, opened 2015-01-24
Fix: Refer to actual view variables

localheinz:
Refer to actual view renderer properties.


module/Application/view/application/index/index.phtml
```diff
@@ -59,7 +59,7 @@
                 </div>
             </div>
         <?php endforeach; ?>
-        <?php echo $this->paginationControl($this->repositories, 'Sliding', 'application/index/pagination', array('query' => $query)) ?>
+        <?php echo $this->paginationControl($this->repositories, 'Sliding', 'application/index/pagination', array('query' => $this->query)) ?>
```
GeeH: Any particular reason @localheinz?

ins0:
@localheinz should i implement this in my pr #333 so we don't get in conflict with each other?

localheinz:
@ins0

:+1:

PR #336, opened 2015-01-24
Enhancement: Leverage LoggerAbstractServiceFactory for creating logger

localheinz:
This PR

* [x] adds a test to assert the service `ApplicationServiceErrorHandling` can be created
* [x] leverages `Zend\Log\LoggerAbstractServiceFactory` for creating logger
* [x] extracts a concrete factory for the `ApplicationServiceErrorHandling` service
* [x] leverages the `class` keyword as a service identifier
* [x] modifies type-hint to interface instead of concretion
* [x] reduces visibility of class property to `private`


module/Application/config/module.config.php
```diff
@@ -85,12 +86,31 @@
...
     ],
+    'log' => [
+        'ZendLog' => [
+            'writers' => [
+                [
+                    'name' => 'stream',
+                    'options' => [
+                        'stream' => sprintf(
+                            './data/logs/log_%s.txt',
+                            date('F')
```
snapshotpl: When config cache is enabled on production this value will stay constant

module/Application/config/module.config.php
```diff
@@ -85,12 +86,31 @@
         ],
     ],
     'service_manager' => [
+        'abstract_factories' => [
+            Log\LoggerAbstractServiceFactory::class,
```
snapshotpl: Why you don't used http://framework.zend.com/manual/current/en/modules/zend.mvc.services.html#zend-log-loggerabstractservicefactory ?

Ocramius: @snapshotpl that is this factory? :)

module/Application/config/module.config.php
localheinz: @snapshotpl 

Good catch.

module/Application/config/module.config.php
snapshotpl: @snapshotpl go home, you are drunk :beers:

module/Application/config/module.config.php
snapshotpl: In monolog exists nice rotating file handler https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/RotatingFileHandler.php
Maybe write something like that for Zend\Log?

Ocramius: I also don't have anything against just using monolog, heh.

localheinz:
@Ocramius @snapshotpl 

Working on integrating monolog, hang on!

Ocramius:
@localheinz nice!

PR #339, opened 2015-01-26
HOTFIX: changed github api call order

ins0:
as explained here https://github.com/EvanDotPro/EdpGithub/issues/23

the markdown request (POST) needs to be called last as it destroys the normal requests


Ocramius:
Weird bug. Merging as there is no meaningful way to test this here: should be fixed in EdpGithub instead.

ins0:
thanks @Ocramius - so fast that the travis build failed :speedboat:

Ocramius:
Grey builds all the things

ins0:
:smiling_imp:

Issue #251, opened 2014-12-06
Unknown column 'state' in 'field list'


localheinz:
Following https://github.com/zendframework//pull/250 I tried to sign in into the app, which resulted in the following exception being displayed:

### Exception

`Zend\Db\Adapter\Exception\InvalidQueryException`

### File

`vendor/zendframework/zendframework/library/Zend/Db/Adapter/Driver/Pdo/Statement.php:244`

### Message

```
Statement could not be executed (42S22 - 1054 - Unknown column 'state' in 'field list')
```

### Stack Trace

```
#0 vendor/zf-commons/zfc-base/src/ZfcBase/Mapper/AbstractDbMapper.php(141): Zend\Db\Adapter\Driver\Pdo\Statement->execute()
#1 vendor/zf-commons/zfc-user/src/ZfcUser/Mapper/User.php(55): ZfcBase\Mapper\AbstractDbMapper->insert(Object(User\Entity\User), NULL, NULL)
#2 vendor/socalnick/scn-social-auth/src/ScnSocialAuth/Authentication/Adapter/HybridAuth.php(534): ZfcUser\Mapper\User->insert(Object(User\Entity\User))
#3 vendor/socalnick/scn-social-auth/src/ScnSocialAuth/Authentication/Adapter/HybridAuth.php(450): ScnSocialAuth\Authentication\Adapter\HybridAuth->insert(Object(User\Entity\User), 'github', Object(Hybrid_User_Profile))
#4 vendor/socalnick/scn-social-auth/src/ScnSocialAuth/Authentication/Adapter/HybridAuth.php(110): ScnSocialAuth\Authentication\Adapter\HybridAuth->githubToLocalUser(Object(Hybrid_User_Profile))
#5 [internal function]: ScnSocialAuth\Authentication\Adapter\HybridAuth->authenticate(Object(ZfcUser\Authentication\Adapter\AdapterChainEvent))
#6 vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php(468): call_user_func(Array, Object(ZfcUser\Authentication\Adapter\AdapterChainEvent))
#7 vendor/zendframework/zendframework/library/Zend/EventManager/EventManager.php(207): Zend\EventManager\EventManager->triggerListeners('authenticate', Object(ZfcUser\Authentication\Adapter\AdapterChainEvent), Object(Closure))
[...]

PR #252, opened 2014-12-06
Enhancement: Use Doctrine migrations for database schema changes

localheinz:
This PR

* [x] adds `doctrine/orm-module`, `doctrine/migrations` and `symfony/console` as dependencies (with versions that don't break the console commands
* [x] enables `DoctrineModule` and `DoctrineORMModule`
* [x] adds default configuration to `conf/autoload/global.php` and placeholder configuration in `conf/autoload/database.local.php.dist`
* [x] splits the existing SQL dump into two migrations, the first one reflecting the state on the production server, the second one applying the necessary changes to fix #251
* [x] updates `README.md`

PR #340, opened 2015-01-27
Fix: added auto_increment to sql dump

### LINKED ISSUE #306: adding a module
I cloned this project from git and set up a db and git keys. 
Now I try to add a module from Github and get a sql error.
"Statement could not be executed (23000 - 1048 - Column 'module_id' cannot be null)"

This is an insert statement and I believe module_id column should be auto-increment. This isn't the case if you use the .sql file in the git. 

The module mapper's insert looks like this. What's up??
 $result = parent::insert($entity, $tableName, $hydrator);
        $entity->setId($result->getGeneratedValue());
        return $result;

ins0:
fixes #306 

and #251 can be closed?

### DIFF
--- a/data/sql/0.sql
+++ b/data/sql/0.sql
@@ -9,7 +9,7 @@
 --
 
 CREATE TABLE IF NOT EXISTS `module` (
-`module_id` int(11) NOT NULL,
+  `module_id` int(11) NOT NULL AUTO_INCREMENT,
   `name` varchar(255) NOT NULL,
   `description` text NOT NULL,
   `url` varchar(500) NOT NULL,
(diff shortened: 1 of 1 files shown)


localheinz:
@ins0 

Looks good to me, but since @GeeH deploys, I think he should have a look!

localheinz:
@GeeH 

If we merge this, how are you going to make the changes to the database? Manually? Since we decided against #252, shall we maybe use http://phinx.org?

localheinz:
@ins0 

I'm happy to merge this so I could remove this workaround in a test

* https://github.com/localheinz/modules.zendframework.com/blob/feature/search/module/ZfModule/test/ZfModuleTest/Integration/Mapper/ModuleTest.php#L176-L191

however, my concerns in https://github.com/zendframework/modules.zendframework.com/pull/340#issuecomment-73226042 are still valid, no?

ins0:
as the page was working (adding/removing) repositorys and the ``state`` bug was fixed too before this PR, the database should be allready in this format and could be safely merged. no?

localheinz:
@ins0 

I have no idea, I don't have access to the production system.

@GeeH 

Can you confirm that the @ins0 statements are true for the production system?

localheinz:
Confirming that the database schema in production is up-to-date with this, and merging.

Thank you, @ins0!

ins0:
i love your server access :dancers:  - thanks!

gianarb:
giphy

localheinz:
@gianarb 

:smile:

PR #345, opened 2015-01-27
Drop AssetManager module from dev-mode

adamlundrigan:
Drop AssetManager module from dev-mode modules as it's loaded already in application.config.php


localheinz:
Good catch, @adamlundrigan, and thank you!

PR #346, opened 2015-01-27
Feature: added composer to view

ins0:
if we request the composer file allready we should display it.
this fixes also the text display issue with indentation when using the ``pre`` element


localheinz:
@ins0 

Great idea!

Not sure though if I'm the only one who prefers to actually see the `composer.json`,  maybe. Guess we can reiterate, though!

PR #347, opened 2015-01-27
Fix: Changed escape to correct escape attr helper

ins0:
as above + short syntax


adamlundrigan:
:+1:

localheinz:
Thank you, @ins0!

ins0:
thanks @localheinz for this morning merge run

localheinz:
@ins0 

:smile_cat:

PR #348, opened 2015-01-27
Display GitHub error message when module sync fails

adamlundrigan:
Catch the `RuntimeException` thrown by EdpGithub and send it back to the user.

image

Two notes:
 - I tried to test it but failed.  I couldn't find the right incantation to get PHPUnit's mock builder to build a suitable replacement for EdpGithub's RepositoryCollection.  Perhaps someone with a kinder disposition towards mock builder could try?
 - I looked through EdpGithub and GitHub's docs and I don't see any case where sensitive information could be leaked through the failure message, so IMO it should be fine to just dump it back to the user for now.


ins0:
maybe we move the error catch also to the ``RepositoryRetreiver`` class?
https://github.com/zendframework/modules.zendframework.com/blob/master/module/Application/test/ApplicationTest/Service/RepositoryRetrieverTest.php#L201 mocked a error case with the github client - helpfull?

adamlundrigan:
It's a largely cosmetic change anyway so I didn't put too much effort into writing a test for it...that whole controller is untested and needs a refactor to make that feasible.  You're right though, if I mock the client instead of the collection I'd likely have an easier time.

module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -110,12 +120,22 @@ public function organizationAction()
...
+        ));
         $viewModel->setTerminal(true);
         $viewModel->setTemplate('zf-module/index/index.phtml');
+        
+        try {
+            $repos = $this->repositoryRetriever->getUserRepositories($owner, $params);
+            $repositories = $this->fetchModules($repos);
+            $viewModel->setVariable('repositories', $repositories);
+        } catch (GithubException $ex) {
+            $viewModel->setVariable('error_message', $ex->getMessage());
```
localheinz: `camelCase`?

module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -89,11 +90,20 @@ public function indexAction()
             'direction' => 'desc',
         ];
 
-        $repos = $this->repositoryRetriever->getAuthenticatedUserRepositories($params);
-        $repositories = $this->fetchModules($repos);
-
-        $viewModel = new ViewModel(['repositories' => $repositories]);
+        $viewModel = new ViewModel(array(
```
localheinz: Short syntax?

localheinz:
@adamlundrigan 

Not sure about passing through the actual exception message, as you pointed it out, but looks good to me!

adamlundrigan:
@localheinz thanks for the reivew.  I've fixed the items you pointed out. 

As for returning the exception message it's not ideal but it's a good stop-gap to keep the spinner from spinning forever if there is a problem with the sync.  Ideally EdpGithub or the RepositoryRetriever service would throw distinguishable exceptions (or at least exception messages we know to be safe to send back to the user)

module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -89,11 +90,20 @@ public function indexAction()
...
+            'error_message' => '',
+        ]);
         $viewModel->setTerminal(true);
+        
+        try {
+            $repos = $this->repositoryRetriever->getAuthenticatedUserRepositories($params);
+            $repositories = $this->fetchModules($repos);
+            $viewModel->setVariable('repositories', $repositories);
+        } catch (GithubException $ex) {
+            $viewModel->setVariable('errorMessage', $ex->getMessage());
```
Ocramius: You should still set a `503` response code, even if the page is correctly generated.

module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -110,12 +120,22 @@ public function organizationAction()
             'direction' => 'desc',
         ];
 
-        $repos = $this->repositoryRetriever->getUserRepositories($owner, $params);
-        $repositories = $this->fetchModules($repos);
 
-        $viewModel = new ViewModel(['repositories' => $repositories]);
+        $viewModel = new ViewModel([
+            'repositories' => [],
+            'error_message' => '',
```
Ocramius: `error_message` or `errorMessage`?

module/ZfModule/src/ZfModule/Controller/IndexController.php
```diff
@@ -89,11 +90,20 @@ public function indexAction()
             'direction' => 'desc',
         ];
 
-        $repos = $this->repositoryRetriever->getAuthenticatedUserRepositories($params);
-        $repositories = $this->fetchModules($repos);
-
-        $viewModel = new ViewModel(['repositories' => $repositories]);
+        $viewModel = new ViewModel([
+            'repositories' => [],
+            'error_message' => '',
```
Ocramius: `error_message` or `errorMessage`?

module/ZfModule/view/zf-module/index/index.phtml
```diff
@@ -1,4 +1,6 @@
-<?php if (empty($this->repositories)): ?>
+<?php if (!empty($this->errorMessage)): ?>
+    <div class="alert alert-block"><?php echo $this->escapeHtml($this->errorMessage); ?></div>
```
Ocramius: should be red, imo

Ocramius:
I'd still ask for a _unit_ test of this functionality (mock with `->throwsException()`)

module/ZfModule/src/ZfModule/Controller/IndexController.php
adamlundrigan: Probably a good idea, but outside the scope of this PR.  The github sync process needs to be rewritten (#349) and @ins0 is working on the user panel stuff in #351

module/ZfModule/src/ZfModule/Controller/IndexController.php
adamlundrigan: That's what I get for trying to fix it while on my phone :/

module/ZfModule/view/zf-module/index/index.phtml
adamlundrigan: :+1:

adamlundrigan:
@Ocramius I mocked all the things; there are now four test cases covering the changes made in this PR

localheinz:
@adamlundrigan 

Needs a rebase!

Ocramius:
This needs another rebase, sorry for letting it slip through, @adamlundrigan :(

GeeH:
Can this be merged

GeeH:
Ping @Ocramius @adamlundrigan

GeeH:
Please rebase this onto `develop` branch.

Ocramius:
It needs a rebase before it can be merged

adamlundrigan:
I'll take a look today

adamlundrigan:
The code had changed so much I ditched my original and implemented it fresh, integrating your suggestions:

 - `RepositoryRetriever` now catches the exception and returns false (@ins0), which is more consistent with the way other methods in `RepositoryRetriever` behave. 
 - `indexAction` and `listAction` return 503 when an error is detected (@ocramius)
 - The actual exception message is no longer passed through to the user (@localheinz)

I was no longer able to reproduce the original API limit exceptions against GitHub (kudos to whomever fixed the sync :+1:) but the principle behind the change is the same:  catch any exception coming from EdpGithub and dump out an error message to the user.

PR #350, opened 2015-01-28
Enhancement: Use fabpot/php-cs-fixer instead of squizlabs/php_codesniffer

localheinz:
This PR

* [x] removes the dependency on `squizlabs/php_codesniffer`
* [x] requires `fabpot/php-cs-fixer` as a dev dependency
* [x] adds some configuration for `php-cs-fixer` (**please discuss!**)
* [x] performs a dry-run of `php-cs-fixer fix` on Travis
* [x] adds composer scripts to conveniently run the fixer during development
* [x] actually runs the fixer (see 6fb1434)

```
# performs a dry-run
$ composer cs

# fixes violations
$ composer cs-fix
```

Follows https://github.com/zendframework/zf2/pull/6182#issuecomment-68220203 and particularly the overall improvement of `php-cs-fixer`!

h/t @keradus


localheinz:
:bulb: 

Build will fail, of course :trollface: , but when everyone is fine with this, we could actually run the fixer to make it pass.

ins0:
personally i don't like automated code fixes/changes on build - and think it was fair enough just to notice the dev on build - but as the tests run after the fixer i'm fine. :palm_tree:

gianarb:
@ins0 in my project I use php-cs-fix to lint my code into the develop branch after commit and after 4 months I don't have any code problem :+1: In the last version this tool is very stable.. :+1:

ins0:
thanks @gianarb for this nice info :+1:

.php_cs
```diff
@@ -0,0 +1,67 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->exclude('data')
+    ->exclude('puphpet')
+    ->exclude('public/css')
+    ->exclude('public/img')
+    ->exclude('public/js')
+    ->exclude('vendor')
```
keradus: no need to (see https://github.com/keradus/PHP-CS-Fixer/blob/master/Symfony/CS/Finder/DefaultFinder.php#L36)

.php_cs
```diff
@@ -0,0 +1,67 @@
...
+        'short_tag',
+        'spaces_before_semicolon',
+        'spaces_cast',
+        'standardize_not_equal',
+        'ternary_spaces',
+        'trailing_spaces',
+        'unused_use',
+        'visibility',
+        'whitespacy_lines',
+
```
keradus: no need for empty line

.php_cs
```diff
@@ -0,0 +1,67 @@
...
+    ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
+    ->setUsingCache(true)
+    ->fixers([
+        'braces',
+        'concat_with_spaces',
+        'controls_spaces',
+        // 'double_arrow_multiline_whitespaces',
+        'duplicate_semicolon',
+        'elseif',
+        // 'empty_return',
```
keradus: Why is it (and few others) commented?

snapshotpl:
I think also we shouldn't do that. Any changes in sources should be done before commit. In travis we could run test. +1 for php-cs-fixer in require-dev, -1 for run it in travis.

.php_cs
```diff
@@ -0,0 +1,67 @@
...
+    ->exclude('puphpet')
+    ->exclude('public/css')
+    ->exclude('public/img')
+    ->exclude('public/js')
+    ->exclude('vendor')
+    ->in(__DIR__)
+;
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
```
keradus: Maybe use PSR2 level and then set less fixers manually ?

.php_cs
```diff
@@ -0,0 +1,67 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
```
keradus: You may use sth like
```php
$config = new Symfony\CS\Config\Config();
$config->...
$config->getFinder()->exclude(...)->in(__DIR__);

return $config;
```

to not create Finder instance twice

.php_cs
```diff
@@ -0,0 +1,67 @@
...
+    ->exclude('data')
+    ->exclude('puphpet')
+    ->exclude('public/css')
+    ->exclude('public/img')
+    ->exclude('public/js')
```
keradus: btw, if file doesn't have any .php files then you don't need to exclude it

keradus:
> I think also we shouldn't do that. Any changes in sources should be done before commit. In travis
> we could run test. +1 for php-cs-fixer in require-dev, -1 for run it in travis.

It's great to check on travis if user broke the CS rules and told him that.
If not then no-one will remember about it and then there will be a dummy PR only to fix CS after someone miss that.

.php_cs
localheinz: @keradus 

Just figured it made sense to skip searching through these directories, should speed up the process, no?

keradus: It depends how many files are there. But ok, I see ;)

localheinz:
@keradus 

Thanks for the heads-up, incorporated the changes you suggested. 

About the fixers, I'm not sure about which ones everyone else would like to see, so I'm asking for comments and whatnot, please feel free to suggest!

localheinz:
@ins0 

The fixes are not applied automatically, you still have to run

```
$ composer cs-fix
```

yourself, but as pointed out, it helps keeping the code base somewhat clean and also cleaning things up is a lot better with `php-cs-fixer` (although `phpcs` also has `phpcbf`).

ins0:
@localheinz thanks i just overread the ``--dry-run`` option on travis - so revert my first comment here :smile:

keradus:
@ins0 even if there will be no `--dry-run` then Travis works only on it's local copy and won't commit that.

.php_cs
```diff
@@ -0,0 +1,66 @@
...
+;
+
+return $config
+    ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
+    ->setUsingCache(true)
+    ->fixers([
+        'braces',
+        'concat_with_spaces',
+        'controls_spaces',
+        // 'double_arrow_multiline_whitespaces',
```
keradus: :+1: for enabling this

.php_cs
keradus: :+1: for enabling this

.php_cs
```diff
@@ -0,0 +1,66 @@
...
+        'multiline_spaces_before_semicolon',
+        'multiple_use',
+        'namespace_no_leading_whitespace',
+        'new_with_braces',
+        'object_operator',
+        'operators_spaces',
+        'ordered_use',
+        'parenthesis',
+        // 'phpdoc_indent',
+        // 'phpdoc_params',
```
keradus: :+1: for enabling this (both)

ins0:
yeah just realized this, thanks @keradus @localheinz

.php_cs
```diff
@@ -0,0 +1,44 @@
...
+    ->exclude('puphpet')
+    ->in(__DIR__)
+;
+
+return $config
+    ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
+    ->setUsingCache(true)
+    ->fixers([
+        'concat_with_spaces',
+        'controls_spaces',
```
keradus: what is it?

.php_cs
```diff
@@ -0,0 +1,44 @@
...
+        // 'double_arrow_multiline_whitespaces',
+        'duplicate_semicolon',
+        // 'empty_return',
+        'extra_empty_lines',
+        'include',
+        'join_function',
+        'multiline_array_trailing_comma',
+        'multiline_spaces_before_semicolon',
```
keradus: This disallow you to use
```php
$foo
    ->bar()
    ->baz()
;
```

.php_cs
```diff
@@ -0,0 +1,44 @@
...
+        'namespace_no_leading_whitespace',
+        'new_with_braces',
+        'object_operator',
+        'operators_spaces',
+        'ordered_use',
+        // 'phpdoc_indent',
+        // 'phpdoc_params',
+        'return',
```
keradus: almost everything is alphabetic, this one doesn't

keradus:
I would add `no_blank_lines_after_class_opening` and `no_empty_lines_after_phpdocs`.

snapshotpl:
I missed `--dry-run`. Now it make sense! :+1:

composer.json
```diff
@@ -16,12 +16,12 @@
         "ezyang/htmlpurifier": "4.6.*"
     },
     "require-dev": {
+        "bjyoungblood/BjyProfiler": "1.1.0",
```
Ocramius: Do we actually use this thing?

ins0: only in dev mode https://github.com/zendframework/modules.zendframework.com/blob/8af567eaa08f2f67ecf6c0f55998065201ff0cf0/config/autoload/development.local.php.dist#L10

localheinz:
@Ocramius 

Since you and I have some history with this, would you be willing to have a look at this?

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_script:
   - mysql -uroot modules_test < data/sql/0.sql
 
 script:
-  - vendor/bin/phpcs --standard=phpcs.xml -np --report=summary .
+  - vendor/bin/php-cs-fixer fix --config-file=./.php_cs --dry-run -v
```
Ocramius: add a `--diff` here

composer.json
```diff
@@ -39,6 +39,13 @@
             "UserTest": "module/User/test/",
             "ZfModuleTest": "module/ZfModule/test/"
         }
+    },
+    "scripts": {
```
Ocramius: This should be run by `.travis.yml`, not by composer.

localheinz: @Ocramius 

This doesn't actually run, it just allows to run

```
$ composer cs
$ composer cs-fix
```

The equivalent would be a adding a `Makefile`

```Makefile
cs:
	vendor/bin/php-cs-fixer fix --config-file=./.php_cs --dry-run --diff -v

cs-fix:
	vendor/bin/php-cs-fixer fix --config-file=./.php_cs --diff -v
```

and then being able to execute

```
$ make cs
$ make cs-fix
```

keradus: It's just aliasing a more clean `php-cs-fixer fix` commad and map one parameter into another \(`cs`/`cs-fix` vs `--dry-run --diff`)
Personally I can't see any benefits

localheinz: @keradus 

You can't see the benefit of typing

```
$ composer cs
```

vs

```
$ php vendor/bin/php-cs-fixer fix --config-file=./.php_cs --dry-run --diff -v
```

I think the benefit is obvious: fewer characters to type, no?

localheinz: @keradus 

Never mind, I see that `php-cs-fixer` automatically picks up the configuration file - I'm so used to specifying it (I once opened a PR against the original repo which actually allowed to specify the path), that I forgot about this. Hang on a second.

keradus: I jest like to run my local `php-cs-fixer` with my local `.php_cs` like `php-cs-fixer fix` (which is as schort as `composer cs` ;)

module/Application/Module.php
```diff
@@ -6,5 +6,4 @@
  * @copyright Copyright (c) 2005-2012 Zend Technologies USA Inc. (http://www.zend.com)
  * @license   http://framework.zend.com/license/new-bsd New BSD License
  */
-
 require_once __DIR__ . '/src/Application/Module.php';
```
Ocramius: Note that this is not actually required anymore

Ocramius:
@localheinz this needs a rebase

localheinz:
@Ocramius 

Rebased!

Ocramius:
@localheinz le merge \o/

localheinz:
Thank you, @Ocramius!

.php_cs
```diff
@@ -0,0 +1,45 @@
...
+        // 'multiline_spaces_before_semicolon',
+        'namespace_no_leading_whitespace',
+        'new_with_braces',
+        'no_blank_lines_after_class_opening',
+        'no_empty_lines_after_phpdocs',
+        'object_operator',
+        'operators_spaces',
+        'ordered_use',
+        // 'phpdoc_indent',
+        // 'phpdoc_params',
```
keradus: Why merged with a lot of commented fixers?
Please uncomment or delete.

TomasVotruba:
In case of more switching ideas, this might be an inspiration: https://www.tomasvotruba.cz/blog/2017/05/03/combine-power-of-php-code-sniffer-and-php-cs-fixer-in-3-lines/

keradus:
I think one CS tool is enough, whichever it would be

TomasVotruba:
I guess author before this PR though too.

Issue #349, opened 2015-01-28
Strategies to avoid hitting the rate limit?


adamlundrigan:
At the very least we should be caching the result of API calls used to generate the [module view page](https://github.com/zendframework/modules.zendframework.com/blob/master/module/ZfModule/src/ZfModule/Controller/IndexController.php#L56-65).  A better solution might be to follow Packagist's lead: indefinitely cache the initial result during module add, then listen for repository changes via a webhook handler.  Of course, this will require us to instruct the module owner on how to enable it.

ins0:
allready implement as the github client uses etags to avoid increasing the current api limit if a repository content did not changed.

https://github.com/EvanDotPro/EdpGithub/blob/master/src/EdpGithub/Listener/Cache.php#L42

authenticated request to github had a rate limit of ``5000`` calls per hour and ``50`` for unauthenticated requests.

adamlundrigan:
I haven't looked into it too closely yet, but I set up my local dev env, logged in, and tried to sync my modules and after a sizable wait received the "API limit has been exceeded" message.  My GitHub account has 56 repositories.

ins0:
now i got the point, there is one request to collect all the user repositorys but the old function https://github.com/zendframework/modules.zendframework.com/blob/master/module/ZfModule/src/ZfModule/Service/Module.php#L71 will use one call against every single repository to filter only zf2 modules.

i would love to see this check moved to the ``register`` process where the check will be executed.

adamlundrigan:
One optimization could be to use the search instead of loading a list of all the user's repositories.  As an example, with this query: 
```
user:adamlundrigan filename:Module.php "class Module"
```
I get only the repositories that match our criteria for being a module ([see my result](https://api.github.com/search/code?q=user:adamlundrigan%20filename:Module.php%20%22class%20Module%22)).  This also works for organization names.

ins0:
@adamlundrigan awesome :+1: like

adamlundrigan:
Something's fishy with the rate limit on my local clone.  Invoking my awesome powers of [VDD](https://twitter.com/andrewvos/status/174179120610160640) I added this to `EdpGithub\Listener\Cache::postSend`:

```
var_dump($response->getHeaders()->get('X-RateLimit-Remaining')->getFieldValue());
```

And this is the progression of `X-RateLimit-Remaining` when I try to sync my repositories:

[...]

PR #351, opened 2015-01-28
Enhancement: User Panel

ins0:
this pr will

- [x] trigger the modules ajax request only when the user navigates to the correspondending tab
- [x] view fixes
- [x] added more string escaping to unsafe vars


module/User/view/zfc-user/user/index.phtml
```diff
@@ -1,3 +1,17 @@
+<?php
+$this->inlineScript()->appendScript(<<<EOT
```
Ocramius: note that you should use more explicit names, such as `<<<JAVASCRIPT`. Also use nowdoc, not heredoc:

`$this->inlineScript()->appendScript(sprintf($template, $this->escapeJs($this->url('zf-module'))));`

Ocramius:
> trigger the modules ajax request only when the user navigates to the correspondending tab

Is this actually an improvement? I'd rather either have a separate page or still have the content.

ins0:
i would say so as the outcome is the same as before just saves api calls if a user is logged in but don't hit the repository tab. but after the outcome of #349 this will be replaced in a future version and is minor

ins0:
will close as this is outdated and may improved on different way

localheinz:
@ins0 

What do you think, shall we update the actual design of the site as well? I think it could be improved a lot!

ins0:
@localheinz i'm not skilled in creating redesigns/ux design - but why not and give it a try :8ball: yep

localheinz:
@ins0 

:+1:

Ocramius:
:8ball: good luck!

PR #352, opened 2015-01-29
[WIP] [RFC] Implement Packagist-style module add process

adamlundrigan:
## Overview

Drop the current module add process (n+1 GitHub API requests to fetch all the repositories for your account and determine which are ZF2 modules) in favour of something closer to what Packagist does:

 1. Enter the URL to your repository
 2. In an AJAX request, pull the repository and determine if it's a ZF2 module
   - If it is a module, import it and return the new module's details + link to it's view page
   - If it isn't a module (or if something fails), return a suitable error message
 3. If a module was successfully created, add a row for it below the add form so the user can see it has been added and either open it's view page in a new window or continue adding more modules. 

There are two main upsides to this approach
 - Uses far fewer GitHub API requests, which is important as the Search API used to determine if a repository contains a ZF2 module is limited to 30 requests per minute.  
 - Makes supporting multiple sources (eg: bitbucket) easier.

## TODO list

_TODO: needs more TODOs_

 - [ ] Implement UI for triggering module add
 - [ ] Implement background handling of add requests
  - [ ] Packagist package names
  - [ ] GitHub repository slug

PR #353, opened 2015-01-29
List of repositories from Github

gianarb:
Hi! This proposal is not a resolution of all problems but in this moment I think that there are only one target..
We make zf.modules usable..
Login, submit repos in the first step.. This proposal tries to "resolve" the rate limit problem.

See
* #349 

With this PR the list of repos is inconsistente because there are a lot of repos wrong (not zf2 modules).. But works and if you try to submit a wrong repo exists a check for that :smile: 

Now there are already problem if i try to submit a good module I have this expection
`Statement could not be executed (23000 - 1048 - Column 'module_id' cannot be null)`

But I don't remember the story of the `module_id` :calendar: 

This isn't the final resolution :)

### DIFF
--- a/module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
+++ b/module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
@@ -120,15 +120,8 @@ class IndexControllerTest extends AbstractHttpControllerTestCase
         ;
 
         $moduleService
-            ->expects($this->any())
+            ->expects($this->never())
             ->method('isModule')
-            ->willReturnCallback(function ($repository) use ($nonModule) {
-                if ($repository !== $nonModule) {
-                    return true;
-                }
-
-                return false;
-            })
         ;
 
         $moduleMapper = $this->getMockBuilder(Mapper\Module::class)
--- a/module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
+++ b/module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
@@ -280,18 +280,15 @@ class IndexControllerTest extends AbstractHttpControllerTestCase
         $this->assertResponseStatusCode(Http\Response::STATUS_CODE_200);
     }
 
-    public function testOrganizationActionRendersValidModulesOnly()
+    public function testOrganizationActionRendersUnregisteredModulesOnly()
     {
         $this->authenticatedAs(new User());
 
-        $validModule = $this->validModule();
-
-        $nonModule = $this->nonModule();
+        $unregisteredModule = $this->validModule();
         $registeredModule = $this->registeredModule();
 
         $repositories = [
-            $validModule,
-            $nonModule,
+            $unregisteredModule,
             $registeredModule,
             $this->forkedModule(),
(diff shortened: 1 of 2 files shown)


adamlundrigan:
Are you working on this?  I'm about to dig into #352 ([WIP] [RFC] Implement Packagist-style module add process) but I don't want to waste my time if someone else is already doing it

gianarb:
Hi.. This is only another proposal to fix the problem.. This is very easy.. I don't know.. :) No problem for me I close this PR..

gianarb:
Hi! I reopen this PR because in my opinion this is a good hotfix to resolve (in a first moment) the Module Load problem.. :) 

When the @adamlundrigan feature well be ready we can replace it with new flow! :) 

Thanks for your work!

ins0:
@gianarb needs a rebase :smiley:

gianarb:
Ready! :) Thanks @ins0..

Now into "My Modules" step return all my repositories and if you try to load a `$repository->isModule()` == false return exception..

ins0:
@gianarb if the check is removed from the list will the repository still checked on the ``register`` process?

gianarb:
Yes in this moment this check (isModule) run for each repositories when we call list and after submit the single module..

In this PR I remove the first check.. The list comprises all my repositories (not only module) but works :)

localheinz:
@Ocramius 

Hang on, I missed something!

localheinz:
@gianarb 

Looks good to me in order mitigate the API rate limit problem for now, since we still have the check when adding modules, see right here:

https://github.com/zendframework/modules.zendframework.com/blob/master/module/ZfModule/src/ZfModule/Controller/IndexController.php#L189-L194

@Ocramius 

Care to review?

module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
```diff
@@ -328,15 +315,8 @@ public function testOrganizationActionRendersValidModulesOnly()
         ;
 
         $moduleService
```
Ocramius: Is this mock actually needed?

localheinz: @Ocramius 

Only to assert that `ZfModule\Service\Module::isModule()` is never called!

localheinz:
@Ocramius 

The only thing we've got at the moment is a not-so-user-friendly handling of when you're trying to submit a non-module, an exception is thrown.

However, @ins0 is working on that at the moment.

gianarb:
Yes in my opinion this is only a first fix, maybe in the future we can study another flow

module/ZfModule/test/ZfModuleTest/Integration/Controller/IndexControllerTest.php
Ocramius: Yeah, but that's not critical to this particular test, so it should just be skipped

gianarb:
@Ocramius @localheinz have you news about this PR?
Works? It will merge after #421?

localheinz:
@Ocramius 

I think this can go in - but since I contributed here, I'd prefer you to have a look and merge!

Ocramius:
It is a go indeed

PR #354, opened 2015-01-31
Fix: Remove phpmd/phmd

localheinz:
Since there seems to be some consensus on removing `phpmd/phpmd`, this PR removes it.


Ocramius:
@localheinz thanks!

localheinz:
Sure, you're welcome!

PR #357, opened 2015-02-02
Fix: Remove useless files

localheinz:
This PR

* [x] removes leftovers from previous ZF2 versions module loading mechanism

Follows @Ocramius' comment, see [`Zend\ModuleManager\Listener\ModuleResolverListener`](https://github.com/zendframework/zf2/blob/master/library/Zend/ModuleManager/Listener/ModuleResolverListener.php#L23-L34).

Not sure when this was changed?!

### DIFF
--- a/module/Application/Module.php
+++ b/module/Application/Module.php
@@ -1,9 +1,0 @@
-<?php
-/**
- * Zend Framework (http://framework.zend.com/)
- *
- * @link      http://github.com/zendframework/ZendSkeletonModule for the canonical source repository
- * @copyright Copyright (c) 2005-2012 Zend Technologies USA Inc. (http://www.zend.com)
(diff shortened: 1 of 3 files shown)


adamlundrigan:
Speaking of useless files, I think this fits the bill: https://github.com/zendframework/modules.zendframework.com/blob/master/module/Application/src/Application/View/Helper/FlashMessenger.php
ZF2 has a bundled FlashMessenger view helper

localheinz:
@adamlundrigan 

Actually, I had a look and totally wondered why it was even there. To be honest, I have no idea!

adamlundrigan:
My guess is that this site predates the addition of the flash messenger view helper to the framework in 2.1

PR #361, opened 2015-02-04
Fix: Remove useless configuration

localheinz:
This PR

* [x] removes controller manager configuration which is useless, as a `User\Controller\ModuleController` doesn't even exist


Ocramius:
Thanks @localheinz!

Issue #308, opened 2015-01-18
ViewHelpers should not access database layer


localheinz:
Because that really shouldn't happen, should it?

localheinz:
See

* https://github.com/zendframework/modules.zendframework.com/pull/307#discussion_r23134821
* https://github.com/zendframework/modules.zendframework.com/pull/307#discussion_r23134827
* https://github.com/zendframework/modules.zendframework.com/pull/307#discussion_r23134833

GeeH:
There is nothing inherently wrong with a view helper hitting the database via service as explained to me by @Ocramius at length one fine day in January. Thoughts?

GeeH:
ping @ocramius

PR #362, opened 2015-02-04
Fix: View helper should actually be a controller plugin

localheinz:
This PR

* [x] turns a view helper which should be a controller plugin into a controller plugin
* [x] implements an extension of the `ZfcUser\Controller\UserController` in order to set view variables in the view model returned from the `indexAction()`

### DIFF
--- a/module/User/src/User/Controller/UserController.php
+++ b/module/User/src/User/Controller/UserController.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace User\Controller;
+
+use Zend\View\Model\ViewModel;
+use ZfcUser\Controller\Plugin\ZfcUserAuthentication;
+use ZfcUser\Controller\UserController as ZfcUserController;
+
+/**
+ * @method array listModule(array $options)
+ * @method ZfcUserAuthentication zfcUserAuthentication()
+ */
+class UserController extends ZfcUserController
+{
+    public function indexAction()
+    {
+        if (!$this->zfcUserAuthentication()->hasIdentity()) {
+            return $this->redirect()->toRoute(static::ROUTE_LOGIN);
+        }
+
+        $viewModel = new ViewModel([
+            'modules' => $this->listModule([
+                'user' => true,
+            ]),
+        ]);
+
+        $viewModel->setTemplate('zfc-user/user/index');
+
+        return $viewModel;
+    }
+}
--- a/module/User/test/UserTest/Integration/Controller/UserControllerTest.php
+++ b/module/User/test/UserTest/Integration/Controller/UserControllerTest.php
@@ -3,9 +3,16 @@
 namespace UserTest\Integration\Controller;
 
 use ApplicationTest\Integration\Util\Bootstrap;
+use User\Controller;
+use User\Entity\User;
+use User\View\Helper\UserOrganizations;
 use Zend\Authentication\AuthenticationService;
 use Zend\Http;
+use Zend\Mvc;
 use Zend\Test\PHPUnit\Controller\AbstractHttpControllerTestCase;
+use Zend\View;
+use ZfModule\Mvc\Controller\Plugin\ListModule;
+use ZfModule\View\Helper\TotalModules;
 
 /**
  * @coversNothing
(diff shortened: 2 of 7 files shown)


Ocramius:
@localheinz needs a rebase

localheinz:
Rebased, @Ocramius! Thank you for the review.

module/User/view/zfc-user/user/index.phtml
```diff
@@ -23,7 +23,6 @@
         <div class="tab-content">
             <div class="tab-pane active" id="modules">
                 <?php
-                $modules = $this->listModule(['user' => true]);
```
Ocramius: I don't actually agree with this. Why was it moved to a controller plugin? What's the rationale behind it?

localheinz: As far as I can see from what the view helper does:

* it accesses the database layer
* it makes calls to an external API
* it returns an array of data

As far as my experience goes, these are all things I would never expect from a view helper.

Instead, I would see the responsibility of a view helper in accepting some data and preparing it for presentation in the view, for example, by applying some formatting or filling the data into a view template.

Please note that, as far as I can see, I wouldn't even create a controller plugin just yet, as it's only ever used once. Instead, I'd probably leave it to a service. 

This was just the smallest refactoring possible, I believe.

localheinz:
@Ocramius 

Also see #308.

module/User/view/zfc-user/user/index.phtml
Ocramius: > these are all things I would never expect from a view helper

I disagree with this, but I also see why you'd think that. As long as these are all read operations, IMO it's fine to have them in view helpers. Yes, they sould not cause changes/events to be triggered.

 > Instead, I'd probably leave it to a service

Possibly a good solution, yes.

module/User/src/User/Controller/UserController.php
```diff
@@ -0,0 +1,31 @@
...
+use ZfcUser\Controller\Plugin\ZfcUserAuthentication;
+use ZfcUser\Controller\UserController as ZfcUserController;
+
+/**
+ * @method array listModule(array $options)
+ * @method ZfcUserAuthentication zfcUserAuthentication()
+ */
+class UserController extends ZfcUserController
+{
+    public function indexAction()
```
Ocramius: This controller has no associated tests

Ocramius:
Once the missing tests are introduced (integration with `@coversNothing` is fine) I can merge. I'm fine with the current approach after discussion.

localheinz:
@Ocramius 

Added a test for the case when the authentication service has no identity.

Tried to add a test for the success case and have almost been mocking myself to death and/or trying to detach listeners, still getting failing tests because of that.

module/User/test/UserTest/Integration/Controller/UserControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+        $serviceManager
+            ->setAllowOverride(true)
+            ->setService(
+                'zfcuser_auth_service',
+                $authenticationService
+            )
+        ;
+
+        $listModule = $this->getMockBuilder(ListModule::class)
```
Ocramius: I think mocking `$listModule` as `function () {...}` is enough

Ocramius: (if it makes your life easier: otherwise, current approach is fine)

localheinz: @Ocramius 

> I think mocking `$listModule` as `function () {...}` is enough

Doesn't work, as `PluginManager` is validating and asking for an implementation of `Zend\Mvc\Controller\Plugin\PluginInterface`. However, I mocked out all of the view helpers giving me a headache, tests should pass now!

Ocramius: Grrr @ `PluginManager`.

Yeah, mocking controllers is a mess

Ocramius:
@localheinz let it be merged!

localheinz:
😊

PR #363, opened 2015-02-04
Fix: Attach ModuleRouteListener only once

localheinz:
See https://github.com/zendframework/modules.zendframework.com/blob/master/module/Application/src/Application/Module.php#L18-L20.


Ocramius:
Merged, thanks!

PR #364, opened 2015-02-04
Fix: Remove PHPMD annotations

localheinz:
This PR

* [x] removes useless suppressions of PHPMD warnings

Follows #354.


Ocramius:
@localheinz thanks!

PR #365, opened 2015-02-05
Fix: Move plugin functionality into appropriate service

localheinz:
This PR

* [x] moves the functionality of the `ZfModule\Mvc\Controller\Plugin\ListModule` controller plugin into `ZfModule\Service\Module` as this seems to be the appropriate service (same dependencies as the plugin)
* [x] creates a factory for `User\Controller\UserController` and injects the the aforementioned service
* [x] creates an alias for the `zfcuser` controller so we can actually use a factory 
* [x] adds a bunch of tests for `Service\Module::listModule()` so we can move on and refactor
* [x] extracts `Service\Module::listAllModules()` and `Service\Module::listUserModules()` from `Service\Module::listModule()`, removes the latter
* [x] renames methods 

Follows #362, especially https://github.com/zendframework/modules.zendframework.com/pull/362#discussion_r24087791.

:bulb: Not sure about renaming methods, but to be honest, I have recently gotten tired of all of this `getFoo()` and `setBar()` stuff, does it make sense?


module/User/config/module.config.php
```diff
@@ -5,9 +5,12 @@
 
 return [
     'controllers' => [
-        'invokables' => [
+        'aliases' => [
```
localheinz: Need to alias, otherwise the configuration of https://github.com/ZF-Commons/ZfcUser/blob/1.0.0/config/module.config.php#L8-L12 kicks in.

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -80,4 +80,37 @@ public function isModule(stdClass $repository)
...
+     * @param array $options array of options
+     * @return array Array of modules
+     */
+    public function listModule($options = null)
+    {
+        //need to fetch top lvl ServiceLocator
+        $user = isset($options['user']) ? $options['user'] : false;
+
+        //limit modules to only user modules
+        if ($user) {
```

While I am at it, I would like to remove the options, since there's only one call to this method (YAGNI), but maybe you prefer a separate PR? Still in the mood for cleaning things before moving on to actually implementing new features.

Ocramius:
@localheinz much better :-)

module/ZfModule/src/ZfModule/Service/Module.php
Ocramius: @localheinz why not just providing a custom method? Cleaner and more expressive, IMO

localheinz:
@Ocramius 

Glad you like it!

module/ZfModule/src/ZfModule/Service/Module.php

Wait till I finish my taxes, will follow up with a separate PR!

localheinz:
@Ocramius 

Added tests for `Service\Module::listModule()` and refactored the code a bit, please have a look!

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -80,4 +81,49 @@ public function isModule(stdClass $repository)
...
+            'DESC'
+        );
+    }
+
+    /**
+     * @return Entity\Module[]
+     */
+    public function currentUserModules()
+    {
+        $repositories = $this->githubClient->api('current_user')->repos([
```
Ocramius: Isn't `current_user` hitting the `modules.zendframework.com` user token's user?

To be honest, I have no idea, haven't had much of a look at the API yet. But I'm sure that @ins0 has. Nonetheless, this should be a separate issue.

I'd assume, though, that this refers to the currently authenticated user, no?

Ocramius: Yes, and that is very weird: the service keeps a reference to the user in the auth service? (you see where I'm going with this...)

ins0: jep ``current_user`` will request the user repositorys from the current linked github account of a user

Give us a hint!

Shall we leverage @ins0 `RepositoryRetriever` here? By the way, someone mentioned on Twitter that using *leverage* when *use* would be sufficient is a thing, it's called douchebaggery. So, shall we use it?

Ocramius: Are you asking if we should use *douchebaggery*? :trollface:

Ocramius: As for the service, the problem is that services should be stateless... Seems to be not the case here?

ins0: don't know if i get this correct, but if so it would be nice if someone can explain what i have done wrong with the service so i can fix this (hopefully) :confounded:

Ocramius: @ins0 the problem is just that the user should be passed down to the service, and then the service should extract the token (by user ID or data) and apply operation that way.


In general, any service doing `$foo->getCurrentThing()` (without params) is very fishy, as it seems to assume that global state is involved.

Thoughts?

Agree, it seems weird what's going on! 

@ins0 

As far as I can tell, you've only done good so far! :wink: 

Shall we tackle this thing in a separate PR? I'm thinking that maybe there should just be one service depending on an instance of `EdpGithub\Client` and that should be the service where you go to if you need something from GitHub. 

What do you think?

Ocramius: Separate PR indeed: I just wanted to point out the problem

:grin:

ins0: so this would be ok?
```php
$userRepoInDb = $this->someService->getUserRepos();
foreach($userRepoInDb as $repository){
        $repoData = $this->repositoryRetreiver->getRepositoryMetaData($repository->getName());
       // do stuff check for etc?
}
```
so the service is not forced to call for the current user`?

Ocramius: `$this->someService->getUserRepos();` would still need the `$user` ;-)

ins0: one more try :smiley: 

```php
$userRepoInDb = $this->someService->getUserRepos($this->identity());
foreach($userRepoInDb as $repository){
        $repoData = $this->repositoryRetreiver->getRepositoryMetaData($identity->getUsername(), $repository->getName());
       // do stuff check for etc?
}
```

Ocramius: Yes, that's the stuff

ins0: thanks for your explanation @Ocramius

localheinz: @Ocramius @ins0 

How about this PR, then, are you fine with it?

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -80,4 +81,49 @@ public function isModule(stdClass $repository)
...
+            'type' => 'all',
+            'per_page' => 100,
+        ]);
+
+        $modules = [];
+
+        array_walk($repositories, function ($repository) use (&$modules) {
```
ins0: @localheinz this works? as i think the github client returns and object and this would fail then

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -80,4 +81,49 @@ public function isModule(stdClass $repository)
...
+            if (true === $repository->fork) {
+                return;
+            }
+
+            if (false === $repository->permissions->push) {
+                return;
+            }
+
+            $module = $this->moduleMapper->findByName($repository->name);
+            if (null === $module) {
```
ins0: this would fail i think as the ``moduleMapper`` returns ``false`` and not ``null``

module/ZfModule/src/ZfModule/Service/Module.php
localheinz: @ins0 

Good catch! I guess it's time @EvanDotPro updates the doc-blocks of the corresponding method, no? Anyway, I updated both test and production code.

ins0:
:+1: looks good!

Ocramius:
@localheinz merged, thanks!

localheinz:
Thank yo for reviewing, @Ocramius and @ins0!

PR #366, opened 2015-02-05
Fix: Entity test

localheinz:
This PR

* [x] reduces visibility of a property in a test to the maximum needed
* [x] adds a doc-block
* [x] prevents repetition
* [x] asserts strict


Ocramius:
@localheinz merged, thanks!

PR #367, opened 2015-02-06
Fix: Misspelled method is actually used

localheinz:
This PR

* [x] fixes an issue where a misspelled method `getCreateAtDateTime()` was actually used

Follows #366.


localheinz:
@Ocramius 

Sorry, my bad!

Ocramius:
@localheinz happens all the time, no biggie :-) Merging!

PR #374, opened 2015-02-08
Fix: Remove unused view helper

localheinz:
This PR

* [x] removes a view helper which is neither used nor registered


localheinz:
In case you didn't notice, this is part of a series of me looking at the existing view helpers.

ins0:
:+1:

Ocramius:
Killed, thanks @localheinz!

PR #377, opened 2015-02-08
Fix: Clean up template maps

localheinz:
This PR

* [x] removes non-existent templates from template maps
* [x] keeps remaining templates sorted by key


Ocramius:
@localheinz thanks!

PR #378, opened 2015-02-09
Enhancement: Extract class for User\Mapper\User factory

localheinz:
This PR

* [x] extracts a concrete factory for `User\Mapper\User`


Ocramius:
Merging, no-brainer, thanks!

PR #382, opened 2015-02-10
Enhancement: Cache dependencies installed with Composer on Travis

localheinz:
This PR

* [x] enables caching of dependencies installed with Composer on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ php:
   - 5.6
   - hhvm
 
+cache:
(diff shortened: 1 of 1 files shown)


Ocramius:
@localheinz nice feature! Where did you dig it out from?

gianarb:
@localheinz I see that this feature is only for [Travis-PRO](http://docs.travis-ci.com/user/caching/) :) Np only information :mag:

PR #383, opened 2015-02-10
Fix: Add/fix/update comments for ZfModule\Entity\Module

localheinz:
This PR

* [x] adds inline comments where we're dealing with instances of `ZfModule\Entity\Module` (helps with finding usages, auto-completion and refactoring in a decent IDE
* [x] removes useless comments from accessors and mutators
* [x] adds method-level comments to accessors and mutators (where missing)


ins0:
looks good :+1:

Ocramius:
Merged, thanks!

PR #384, opened 2015-02-10
Fix: Use alternative syntax in view scripts

localheinz:
This PR

* [x] adds a few semicolons (for consistency)
* [x] uses alternative syntax in view scripts (also for consistency)
* [x] removes an assignment and a condition with effect, but no use
* [x] while at it, escapes a few unescaped strings in the view templates affected above

### DIFF
--- a/module/User/view/scn-social-auth/user/login.phtml
+++ b/module/User/view/scn-social-auth/user/login.phtml
@@ -3,13 +3,11 @@
     <p>
         There is no sign-up form anymore! Just log in with your GitHub account. All your information will be provided by GitHub.
     </p>
-    <?php
-    $socialSignIn = true;
-    foreach ($this->options->getEnabledProviders() as $provider) {
-        if ($socialSignIn) {
-            $socialSignIn = false;
-        }
-        echo '<p>' . $this->socialSignInButton($provider) . '</p>';
-    }
-    ?>
+    <?php $socialSignIn = true; ?>
+    <?php foreach ($this->options->getEnabledProviders() as $provider): ?>
+        <?php if ($socialSignIn): ?>
+            <?php $socialSignIn = false; ?>
+        <?php endif; ?>
+        <p><?php echo $this->socialSignInButton($provider); ?></p>
+    <?php endforeach; ?>
 </div>
--- a/module/User/view/scn-social-auth/user/login.phtml
+++ b/module/User/view/scn-social-auth/user/login.phtml
@@ -3,11 +3,7 @@
     <p>
         There is no sign-up form anymore! Just log in with your GitHub account. All your information will be provided by GitHub.
     </p>
-    <?php $socialSignIn = true; ?>
     <?php foreach ($this->options->getEnabledProviders() as $provider): ?>
-        <?php if ($socialSignIn): ?>
-            <?php $socialSignIn = false; ?>
-        <?php endif; ?>
         <p><?php echo $this->socialSignInButton($provider); ?></p>
     <?php endforeach; ?>
 </div>
--- a/module/User/view/zfc-user/user/index.phtml
+++ b/module/User/view/zfc-user/user/index.phtml
@@ -11,7 +11,7 @@
             </div>
         </div>
         <?php foreach ($this->flashMessenger()->getMessages() as $message): ?>
-            <h3 class="zf-green"><?php echo $message ?></h3>
+            <h3 class="zf-green"><?php echo $message; ?></h3>
         <?php endforeach; ?>
 
         <ul class="nav nav-tabs">
@@ -42,17 +42,17 @@
                 ?>
             </div>
             <div class="tab-pane" id="repositories">
-                <div class="well" style="text-align:center">Synchronizing with GitHub <img src="<?php echo $this->basePath('/img/ajax-loader.gif') ?>" alt="loading" /></div>
+                <div class="well" style="text-align:center">Synchronizing with GitHub <img src="<?php echo $this->basePath('/img/ajax-loader.gif'); ?>" alt="loading" /></div>
             </div>
 
             <div class="tab-pane" id="organizations">
-                <?php echo $this->userOrganizations() ?>
+                <?php echo $this->userOrganizations(); ?>
             </div>
         </div>
     </div>
     <div class="col-md-4">
         <div class="sidebar" style="text-align: center">
-            <h3 style="text-align:center">Hello, <?php echo $this->zfcUserDisplayName() ?>!</h3>
+            <h3 style="text-align:center">Hello, <?php echo $this->zfcUserDisplayName(); ?>!</h3>
             <?php /* @var User\Entity\User $identity */ ?>
             <?php $identity = $this->zfcUserIdentity(); ?>
             <img class="thumbnail" src="<?php echo $identity->getPhotoUrl(); ?>" alt="<?php echo $this->zfcUserDisplayName() ?>"/>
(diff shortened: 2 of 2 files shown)


module/User/view/scn-social-auth/user/login.phtml
```diff
@@ -3,13 +3,11 @@
...
-        if ($socialSignIn) {
-            $socialSignIn = false;
-        }
-        echo '<p>' . $this->socialSignInButton($provider) . '</p>';
-    }
-    ?>
+    <?php $socialSignIn = true; ?>
+    <?php foreach ($this->options->getEnabledProviders() as $provider): ?>
+        <?php if ($socialSignIn): ?>
+            <?php $socialSignIn = false; ?>
```
ins0: someone know for this var is used?

localheinz: Maybe some

magic

?

Removing it in a second!

module/User/view/zfc-user/user/index.phtml
```diff
@@ -22,14 +22,11 @@
...
-                    <div class="alert alert-block">No modules was submitted yet</div>
-                    <?php
-                }
-                foreach ($modules as $module) {
-                    echo $this->moduleView([
+                <?php if (empty($modules)): ?>
+                <div class="alert alert-block">No modules was submitted yet</div>
+                <?php endif; ?>
+                <?php foreach ($modules as $module): ?>
+                <?php echo $this->moduleView([
```
ins0: surrounded with an if else ``if (count($modules)...){ else ...`` we can skip the foreach and is cleaner

module/User/view/zfc-user/user/index.phtml
```diff
@@ -37,22 +34,22 @@
...
-                <?php echo $this->userOrganizations() ?>
+                <?php echo $this->userOrganizations(); ?>
             </div>
         </div>
     </div>
     <div class="col-md-4">
         <div class="sidebar" style="text-align: center">
-            <h3 style="text-align:center">Hello, <?php echo $this->zfcUserDisplayName() ?>!</h3>
-            <img class="thumbnail" src="<?php echo $this->zfcUserIdentity()->getPhotoUrl() ?>" alt="<?php echo $this->zfcUserDisplayName() ?>"/>
+            <h3 style="text-align:center">Hello, <?php echo $this->zfcUserDisplayName(); ?>!</h3>
```
ins0: don't know but is this output safe or need escaped?

module/User/view/zfc-user/user/index.phtml
```diff
@@ -37,22 +34,22 @@
...
+            <img class="thumbnail" src="<?php echo $this->zfcUserIdentity()->getPhotoUrl(); ?>" alt="<?php echo $this->zfcUserDisplayName(); ?>"/>
```
ins0: see above

module/User/view/scn-social-auth/user/login.phtml
localheinz: Looks like this was mostly copied from

```php
<?php
$socialSignIn = true;
foreach ($this->options->getEnabledProviders() as $provider) {
    if ($socialSignIn) {
        echo '<h1>Social Sign In</h1>';
        $socialSignIn = false;
    }
    echo '<dd>' . $this->socialSignInButton($provider, $this->redirect) . '</dd>';
}
if ($this->options->getSocialLoginOnly() == false) {
    echo $this->zfcUserLogin;
}
```

See https://github.com/SocalNick/ScnSocialAuth/blob/master/view/scn-social-auth/user/login.phtml#L1-L12.

ins0: looks like its safe to

bye

module/User/view/zfc-user/user/index.phtml
localheinz: I'd escape it, but probably out of the scope of this PR. However, will fix it right away! :wink:

localheinz:
@ins0 

Thank you for reviewing, implemented the suggested changes!

Ocramius:
This requires a rebase

localheinz:
@Ocramius 

Rebased!

Thanks a lot for reviewing everything!

module/User/view/zfc-user/user/index.phtml
```diff
@@ -22,37 +22,34 @@
         <br />
         <div class="tab-content">
             <div class="tab-pane active" id="modules">
-                <?php
-                if (empty($modules)) {
+            <?php if (0 === count($modules)): ?>
+                <div class="alert alert-block">No modules have been submitted yet</div>
+            <?php else: ?>
+                <?php foreach ($modules as $module): ?>
+                    <?php echo $this->moduleView([
```
Ocramius: Why does this helper require so many parameters? Isn't it easier to pass in `$module` instead?

module/User/view/zfc-user/user/index.phtml
```diff
@@ -60,12 +57,4 @@
...
-
-<?php
-$url = $this->url('zf-module');
-$this->inlineScript()->appendScript(<<<EOT
-
-$("#repositories").load("$url");
-
-EOT
-);
+<?php $this->inlineScript()->appendScript('$("#repositories").load("' . $this->url('zf-module') . '");'); ?>
```
Ocramius: Should use `$this->escapeJs()` here

module/User/view/zfc-user/user/index.phtml
```diff
@@ -11,7 +11,7 @@
             </div>
         </div>
         <?php foreach ($this->flashMessenger()->getMessages() as $message): ?>
-            <h3 class="zf-green"><?php echo $message ?></h3>
+            <h3 class="zf-green"><?php echo $message; ?></h3>
```
Ocramius: Should `$this->escapeHtml()` probably, or use the flash messenger directly.

module/User/view/zfc-user/user/index.phtml
localheinz: @Ocramius 

That's what I plan to resolve next! :wink:

module/User/view/zfc-user/user/index.phtml
localheinz: @Ocramius 

Will fix tomorrow morning! Have a good night!

module/User/view/zfc-user/user/index.phtml
```diff
@@ -11,7 +11,7 @@
             </div>
         </div>
         <?php foreach ($this->flashMessenger()->getMessages() as $message): ?>
-            <h3 class="zf-green"><?php echo $message ?></h3>
+            <h3 class="zf-green"><?php echo $this->escapeHtml($message); ?></h3>
```
localheinz: @Ocramius 

Escaping for now, would like to re-iterate on this and use the `FlashMessenger` directly, as suggested! Also, I would like to use http://getbootstrap.com/components/#alerts, what do you think?

Ocramius: Can be done in a different MR then (with screenshots 'n stuff, so someone with better taste than me can deal with it :-) )

ins0: would like to do this if @localheinz agrees - as i allready implemented this on a diff project (personal) :smirk_cat: :

localheinz: @ins0 

Sure, go ahead!

localheinz: @ins0 

If you do, please have a look at places where classes are missing:

```
<div class="alert alert-block">No modules found</div>
```

which results in 

screen shot 2015-02-17 at 16 16 38

vs.

```
<div class="alert alert-info alert-block">No modules found</div>
```

which would result in 
 
screen shot 2015-02-17 at 16 16 56

:bulb: Just as an idea, in [**Getting Real**](https://basecamp.com/books/Getting%20Real.pdf) from 37signals, there's a chapter *The Blank State* - I think we could improve on that.

PR #386, opened 2015-02-10
Enhancement: Add tests for ZfModule\Entity\User

localheinz:
This PR

* [x] adds tests for `ZfModule\Entity\User`
* [x] reduces visibility of fields
* [x] updates comments
* [x] adds inline comments where there is an instance of `ZfModule\Entity\User` in view scripts


Ocramius:
Merged, thanks!

PR #387, opened 2015-02-10
Fix: Remove useless User\Mapper\UserHydrator

localheinz:
This PR

* [x] removes the - what I believe to be useless (unless I'm missing something) - `User\Mapper\UserHydrator`


Ocramius:
Merged, thanks!

PR #389, opened 2015-02-15
Enhancement: Assert InvalidArgumentExceptions are thrown

localheinz:
This PR

* [x] increases coverage for `ZfModule\Mapper\ModuleHydrator` by asserting that an `InvalidArgumentException` is thrown when objects are passed to `hydrate()` or `extract()` which do not implement `ZfModule\Entity\ModuleEntityInterface`


Ocramius:
@localheinz thanks!

PR #391, opened 2015-02-17
Fix: Remove method which is never used

localheinz:
This PR

* [x] removes `ZfModule\Mapper\ModuleInterface::findById()` and its only implementation, as it's never used

:bulb: Not so sure what we'd need `ZfModule\Mapper\ModuleInterface` for, there's currently only one implementation anyway.


Ocramius:
@localheinz merged, thanks!

PR #392, opened 2015-02-17
Fix: Remove plugin configuration for non-existent plugin

localheinz:
This PR

* [x] removes configuration that registers a non-existent controller plugin


Ocramius:
Thanks!

localheinz:
Thank you, @Ocramius!

PR #394, opened 2015-02-18
Enhancement: Also search by owner, split query into words

### LINKED ISSUE #153: Search doesn't work good enough given it's the only way to find things
When I search for a module with the module name di-wrapper using the search string "di wrapper" I don't find anything. Given the fact that the search is the only way of finding things, this doesn't seem sufficient.

localheinz:
This PR

* [x] cleans up `ZfModule\Mapper\Module` a bit
* [x] improves the search by not only looking for matches in `name` and `description`, but also in `owner`
* [x] improves the search by exploding the search query by ` ` and using the parts for searching

Fixes #153.

### Before

screen shot 2015-02-18 at 10 34 59

### After

screen shot 2015-02-18 at 15 50 43

## Before

screen shot 2015-02-18 at 15 22 18

## After

screen shot 2015-02-18 at 15 49 50

### Before

screen shot 2015-02-18 at 15 44 55

### After

screen shot 2015-02-18 at 15 43 35

### DIFF
--- a/module/Application/view/application/index/index.phtml
+++ b/module/Application/view/application/index/index.phtml
@@ -22,10 +22,9 @@
                         <div class="col-xs-7 col-sm-6">
                             <p>
                                 <a href="<?php echo $this->url('view-module', ['vendor' => $this->escapeUrl($module->getOwner()), 'module' => $this->escapeUrl($module->getName())]) ?>">
-                                    <strong><?php echo $this->escapeHtml($module->getName()) ?></strong>
+                                    <strong><?php echo $this->escapeHtml($module->getOwner()); ?>/<?php echo $this->escapeHtml($module->getName()); ?></strong>
                                 </a>
                             </p>
-                            <p><span class="zf-green">Submitter:</span> <?php echo $this->escapeHtml($module->getOwner()) ?></p>
                             <p><span class="zf-green">Created:</span> <?php echo $this->dateFormat($module->getCreatedAtDateTime(), IntlDateFormatter::SHORT, IntlDateFormatter::SHORT); ?></p>
                         </div>
                         <div class="col-xs-4">
--- a/module/ZfModule/src/ZfModule/Mapper/Module.php
+++ b/module/ZfModule/src/ZfModule/Mapper/Module.php
@@ -31,13 +31,9 @@ class Module extends AbstractDbMapper implements ModuleInterface
         }
 
         if (null !== $query) {
-            $select->where(function ($where) use ($query) {
-                $like = '%' . $query . '%';
-
-                /* @var Sql\Where $where */
-                $where->like('name', $like)->or->like('description', $like);
-            });
+            $this->whereLike($select, $query);
         }
+
         $resultSet = new HydratingResultSet($this->getHydrator(), $this->getEntityPrototype());
 
         $adapter = new DbSelect($select, $this->getSql(), $resultSet);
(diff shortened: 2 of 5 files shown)


module/ZfModule/src/ZfModule/Mapper/Module.php
```diff
@@ -77,17 +76,34 @@ public function findByLike($query, $limit = null, $orderBy = null, $sort = 'ASC'
...
+     */
+    private function whereLike(Sql\Select $select, $query)
+    {
+        $select->where(function ($where) use ($query) {
+            $like = '%' . $query . '%';
+
+            /* @var Sql\Where $where */
+            $where
+                ->like('name', $like)
+                ->or
```
Ocramius: urgh

(Yes, I know that's how you do it with Zend\Db :-( )

gianarb: ahah migration to doctrine in this moment is unthinkable :P

localheinz: @Ocramius 

Harhar! Yes, it's ugly.

@gianarb 

Happy to migrate to Doctrine!

Things we could benefit from, then:

* Doctrine migrations
* combination of [`breerly/factory-girl-php`](https://github.com/breerly/factory-girl-php) and [`fzaninotto/Faker`](https://github.com/fzaninotto/Faker) for integration tests

markushausammann:
nice!

markushausammann:
But actually di wrapper should show the package di-wrapper, which it still doesn't seem to do according to the screenshot.

localheinz:
@markushausammann 

Not sure if it's there, hang on a minute.

localheinz:
@markushausammann 

I didn't have it in my local copy of the database, sorry!

localheinz:
Also made some adjustments to the way modules are displayed. I think it makes a lot more sense to always display vendor and package, not just the package name.

markushausammann:
Ok, sounds good, looking forward to seeing it live.

module/Application/view/application/index/index.phtml
```diff
@@ -22,10 +22,9 @@
                         <div class="col-xs-7 col-sm-6">
                             <p>
                                 <a href="<?php echo $this->url('view-module', ['vendor' => $this->escapeUrl($module->getOwner()), 'module' => $this->escapeUrl($module->getName())]) ?>">
-                                    <strong><?php echo $this->escapeHtml($module->getName()) ?></strong>
+                                    <strong><?php echo $this->escapeHtml($module->getOwner()); ?>/<?php echo $this->escapeHtml($module->getName()); ?></strong>
```
snapshotpl: moduleName helper ?

localheinz: @snapshotpl 

Actually, what I would like to do would be a simplification of displaying an item, this definitely shouldn't be repeated all over the place! 

You know, a module should probably have no more than two public views

* side bar
* results (and the live-search thingy)

However, in reality I would even want to go as far as to have a search similar to what all of you know from packagist, what do you think?

Ocramius:
Needs a rebase

localheinz:
@Ocramius 

Rebased, waiting for the build to pass.

Ocramius:
@localheinz awesome, merged! \o/

localheinz:
Thanks a lot, @Ocramius!

I'm very happy to move forward with this thing!

PR #395, opened 2015-02-18
Enhancement: Use class keyword as service identifier in ZfModule module

localheinz:
This PR

* [x] asserts that services can be retrieved
* [x] uses the `class` keyword as service identifier for these services


localheinz:
@Ocramius 

Not really blocking #394, but would be happy to see it there (want to add tests for the mapper).

Ocramius:
Merged, thanks!

PR #397, opened 2015-02-18
Enhancement: Streamline layouts and fix-up content

localheinz:
The aim of this PR is to streamline the existing layouts in order to reduce duplication, so we can move on and actually change it without having to do in two places.

* [x] splits setting up and rendering view helpers
* [x] fixes some alternative syntax blocks spanning multiple lines
* [x] fixes wrapping, indentation, adds missing trailing semicolons (for consistency)
* [x] fixes a closing tag that isn't one
* [x] uses HTML5 style for self-enclosing tags
* [x] removes text that states the obvious
* [x] splits text into paragraphs

### Before

screen shot 2015-02-18 at 17 31 11

### After

screen shot 2015-02-18 at 17 30 52

### DIFF
--- a/module/Application/view/layout/layout-small-header.phtml
+++ b/module/Application/view/layout/layout-small-header.phtml
@@ -35,7 +35,7 @@
                         <?php $identity = $this->zfcUserIdentity(); ?>
                         <div class="gravatar-container">
                             <a class="login" href="<?php echo $this->url('zfcuser'); ?>" title="My profile">
-                                <img src="<?php echo $this->escapeHtmlAttr($identity->getPhotoUrl()); ?>" alt="<?php echo $this->escapeHtmlAttr($this->zfcUserDisplayName()); ?>" style="width:23px;height:23px;" />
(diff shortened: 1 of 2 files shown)


localheinz:
Actually, I'd be very happy to reiterate on this!

Since @ins0 has updated to Twitter Bootstrap 3, we should use it as much as we can. The current design looks too old. 

Also, I suggest to use icons, for example, [Font Awesome](http://fortawesome.github.io/Font-Awesome/)

module/Application/view/layout/layout-small-header.phtml
```diff
@@ -2,53 +2,53 @@
 
 <html lang="en">
     <head>
-        <?php echo $this->headTitle('ZF2 Modules')->setSeparator(' - ')->setAutoEscape(false) ?>
+        <?php $this->headTitle('ZF2 Modules')->setSeparator(' - ')->setAutoEscape(false); ?>
+        <?php echo $this->headTitle(); ?>
```
Ocramius: I liked the previous approach better (easier to read, imo)

Ocramius:
Merged, thanks

PR #400, opened 2015-02-19
Fix: Only load live-search resources when needed

localheinz:
This PR

* [x] moves loading of resources related to live-search to the only place where they are actually needed, which is, the corresponding view script

:bulb: The aim of this PR is to make the diff between `layout/layout.phtml` and `layout/layout-small-header.phtml` smaller, in order to eventually get rid of a lot of duplication.

Follows #397.


Ocramius:
Requires a rebase

localheinz:
@Ocramius 

Rebased! Thank you for reviewing!

Ocramius:
Merged, thanks!

PR #401, opened 2015-02-19
Fix: BasePath helper accepts argument, prepends with /

localheinz:
This PR

* [x] makes use of the fact that the `BasePath` view helper actually accepts an argument, the name of the file to be prepended with the actual base path

Follows f1694ca.


Ocramius:
:+1:

PR #403, opened 2015-02-19
Fix: Use just one layout and switch header based on view variable

localheinz:
This PR

* [x] removes significant duplication by switching rendered headers based on a view variable (only the start page is using the large header anyway) instead of switching the whole layout by module
* [x] removes `EdpModuleLayouts`

Follows  #400.


localheinz:
Needs a rebase onto `master` once #400 is merged.

module/Application/config/module.config.php
```diff
@@ -121,7 +121,7 @@
             'application/index/index'           => __DIR__ . '/../view/application/index/index.phtml',
             'application/index/pagination'      => __DIR__ . '/../view/application/index/pagination.phtml',
             'application/search/index'          => __DIR__ . '/../view/application/search/index.phtml',
-            'layout/error'                      => __DIR__ . '/../view/layout/layout-small-header.phtml',
+            'layout/error'                      => __DIR__ . '/../view/layout/layout.phtml',
```
localheinz: Not sure if the alias `layout/error` is even used anywhere. Would anyone know?

module/Application/view/application/index/index.phtml
```diff
@@ -1,6 +1,10 @@
 <?php $this->headTitle(sprintf('Browse %d modules for Zend Framework 2', $this->totalModules())); ?>
+<?php $this->headLink()->appendStylesheet($this->basePath('css/jquery.livesearch.css')); ?>
+<?php $this->inlineScript()->appendFile($this->basePath('js/jquery.livesearch.js')); ?>
 <?php $this->inlineScript()->appendScript('$(\'#live-search input[name="query"]\').liveSearch({url: "' . $this->url('live-search', [], ['query' => ['query' => '']]) . '"});'); ?>
 
+<?php $this->layout()->withLargeHeader = true; ?>
```
localheinz: Could probably extract a view helper out of this?

```php
<?php $this->pageHeader()->large(); ?>
```

and then in the layout

```php
<?php echo $this->pageHeader(); ?>
```

```php
class PageHeader
{

    private $withSmallHeader = true;

    public function __invoke()
    {
        return $this;
    }

    public function __toString()
    {
        if (true === $this->withSmallHeader) {
            return $this->getView()->render('layout/header-small.phtml');
        }

        return $this->getView()->render('layout/header-large.phtml');
   }
}

**Comment by localheinz** [comment]
@Ocramius 

Just as an FYI for reviewing, as of c02e58a

```
$ cd module/Application/view/layout
$ diff layout-small-header.phtml layout.phtml
```

```diff
57c57
<         <div class="container-fluid landing-intro small-header">
---
>         <div class="container-fluid landing-intro">
60,64c60,69
<                     <div class="col-md-8">
<                         <a href="<?php echo $this->url('home'); ?>" style="text-decoration:none">
<                             <img src="<?php echo $this->basePath('img/zend-logo_small.png'); ?>" alt="ZF Modules">
<                             <p class="zf-green">
<                                 Modules
---
>                     <div class="col-md-8 col-sm-12">
>                         <h1 class="zf-green">
>                             <strong>Welcome to the ZF2 Modules Site</strong>
>                         </h1>
>                         <h3 class="text-muted">Find one of <strong><?php echo $this->totalModules(); ?></strong> modules that fits your needs.</h3>
>                         <div>
>                             <p>
>                                 If you are interested in helping with this project, simply fork the code on
>                                 <a class="zf-green" href="https://github.com/zendframework/modules.zendframework.com" target="_blank">GitHub</a>
>                                 and get started!
66c71,80
<                         </a>
---
>                             <p>
>                                 Additionally, join us in #zftalk.modules on Freenode and ask how you can help!
>                             </p>
>                             <p>
>                                 Grab latest modules using <a class="zf-green" href="<?php echo $this->url('feed') ?>">RSS Feed</a>.
>                             </p>
>                         </div>
>                     </div>
>                     <div class="col-md-4 hidden-xs hidden-sm">
>                         <img class="attentionimage pull-right" src="<?php echo $this->basePath('img/rsz_logo.png'); ?>" alt="ZF Modules">
```

module/Application/view/application/index/index.phtml
Ocramius: Confused by this assignment tbh: what is it doing?

module/Application/view/application/index/index.phtml
```diff
@@ -3,6 +3,8 @@
 <?php $this->inlineScript()->appendFile($this->basePath('js/jquery.livesearch.js')); ?>
 <?php $this->inlineScript()->appendScript('$(\'#live-search input[name="query"]\').liveSearch({url: "' . $this->url('live-search', [], ['query' => ['query' => '']]) . '"});'); ?>
 
+<?php $this->layout()->withLargeHeader = true; ?>
```
localheinz: @Ocramius 

Since I rebased, the comment is outdated, here's the link: https://github.com/zendframework/modules.zendframework.com/pull/403#discussion_r25031345

The idea is this: 

* currently, the only difference between the two layouts is the jumbotron-like header
* the large header is effectively only shown on the start page, that is, this view
* the small header is shown everywhere else
* instead of having two layouts where only the header differs, why not just have one and render the header depending on a view variable?

What do you think?

Ocramius: The idea is good, but is this assigning `withLargeHeader` to the view model?

localheinz: @Ocramius 

As far as I can see from `Zend\View\Helper\Layout`:

```php
public function __invoke($template = null)
{
    if (null === $template) {
        return $this->getRoot();
    }

    return $this->setTemplate($template);
}
```

it should, and it actually does!

localheinz: @Ocramius 

However, if you prefer, I can implement the view helper as suggested in the comment!

Ocramius: I'm fine with the change then.

Ocramius:
:+1:

localheinz:
Thank you, @Ocramius!

PR #405, opened 2015-02-20
Fix: Removed hint

ins0:
as the pr is merged https://github.com/EvanDotPro/EdpGithub/issues/23 steps from  #339 are no longer necessary


localheinz:
@ins0 

Would that need an update of the corresponding dependency, too?

ins0:
yes!

localheinz:
Thanks, @ins0!

PR #407, opened 2015-02-20
Feature: Removed fluid interface from module entity/interface

ins0:
- added missing doctypes to the module interface
- removes return types from interface
- removes fluid style from module entity

### DIFF
--- a/module/ZfModule/src/ZfModule/Entity/ModuleInterface.php
+++ b/module/ZfModule/src/ZfModule/Entity/ModuleInterface.php
@@ -15,7 +15,6 @@ interface ModuleInterface
      * Set Owner
      *
      * @param string
-     * @return string
      */
     public function setOwner($owner);
 
@@ -30,7 +29,6 @@ interface ModuleInterface
      * Set Photo URL
      *
      * @param string
-     * @return string
      */
     public function setPhotoUrl($photoUrl);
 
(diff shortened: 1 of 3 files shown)


module/ZfModule/src/ZfModule/Entity/ModuleInterface.php
```diff
@@ -6,12 +6,16 @@
...
+     *
      * @return string
      */
     public function getOwner();
 
     /**
      * Set Owner
+     *
+     * @param string
+     * @return string
```
localheinz: `@return self`?

Or maybe even nothing? Not sure, but I remember @Ocramius having some arguments against fluent interfaces some time ago!

ins0: you're right :+1: for removing the fluid interface

localheinz: @ins0 

For reference, here's @Ocramius blog post: http://ocramius.github.io/blog/fluent-interfaces-are-evil/

ins0: yeah was reading this after it was published - thanks. may should i extend the pr to remove the fluid style?

localheinz: I think it's just ok to remove the `@return` from the interface, probably!

ins0:
@Ocramius sry this switched from a Fix to a Feature :>

PR #411, opened 2015-02-20
Fix: Wrong CSS class for labels

lku:
There is no `label-important` CSS class in [Bootstrap 3](http://getbootstrap.com/components/#labels) (was in [Bootstrap 2](http://getbootstrap.com/2.3.2/components.html#labels-badges)). Equivalent class in new version is `label-danger`.


Ocramius:
@lku thanks, merged!

PR #413, opened 2015-02-23
Fix: intl extension is required

localheinz:
This PR

* [x] requires `ext-intl`, since it is

See https://github.com/zendframework/modules.zendframework.com/search?utf8=✓&q=IntlDateFormatter.


Ocramius:
@localheinz merged, thanks!

PR #414, opened 2015-02-23
Add feed mapper

snapshotpl:


module/Application/src/Application/Controller/IndexController.php
```diff
@@ -85,4 +73,14 @@ public function feedAction()
 
         return $feedmodel;
     }
+
+    /**
+     * @param int $page
+     * @param string $query
+     * @return \Zend\Paginator\Paginator
+     */
+    protected function getModules($page = 1, $query = null)
```
ins0: i think this method is not necessary and the name don't indicates there actually purpose - there is no need to wrap it into a sep. method

ins0:
@snapshotpl can u please provide tests for this feature?

snapshotpl:
@ins0 Test provided, but Travis looks dead ;)

module/Application/src/Application/Controller/IndexController.php
```diff
@@ -52,32 +56,17 @@ public function feedAction()
         // Prepare the feed
         $feed = new Feed();
         $feed->setTitle('ZF2 Modules');
-        $feed->setDescription('Recently added modules.');
+        $feed->setDescription('Recently added ZF2 modules');
         $feed->setFeedLink('http://modules.zendframework.com/feed', 'atom');
         $feed->setLink('http://modules.zendframework.com');
```
ins0: could you please update the url too? ``zfmodules.com``

module/Application/src/Application/Controller/IndexController.php
```diff
@@ -52,32 +56,17 @@ public function feedAction()
         // Prepare the feed
         $feed = new Feed();
         $feed->setTitle('ZF2 Modules');
-        $feed->setDescription('Recently added modules.');
+        $feed->setDescription('Recently added ZF2 modules');
```
ins0: could you please update the url too? ``zfmodules.com``

module/Application/src/Application/Controller/IndexController.php
ins0: could you please update the url too? ``zfmodules.com``

snapshotpl: I think we should create a new PR with this update in whole project

ins0:
thanks @snapshotpl for providing tests :+1: - if the url from the comment above gets updated this looks good from my point of view :smiley:

module/Application/src/Application/Controller/IndexController.php
ins0: yeah sounds like a better plan :100:

ins0: actually this is the only file that contains the url and needs an update ^^

snapshotpl: There is also link in README

ins0: missed that thanks :+1:

snapshotpl: Or, we should use here serverUrl helper to be more elastic, what are you think?

gianarb: Maybe `serverUrl` is a best solution

ins0: Isn't this a View Helper?

gianarb: mmm you are in Controller..
You can use
```php
$url = $this->url()->fromRoute('home', [], ['force_canonical', true]);
```
@ins0 prints the url into the application in my opinion is not a good idea.. But I don't remember if this solution works

ins0: @gianarb :+1:

snapshotpl: Change in progress...

snapshotpl:
@ins0 @gianarb Done. I extended also feed entry data

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @var Url
+     */
+    protected $url;
+
+    /**
+     * @param Feed $feed
+     */
+    public function __construct(Feed $feed, Url $url)
```
ins0: i would suggest to name this more explicit and not simple ``$url``

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
```diff
@@ -0,0 +1,78 @@
...
+        $moduleName = $module->getName();
+        $urlParams = ['vendor' => $module->getOwner(), 'module' => $moduleName];
+
+        $entry = $this->feed->createEntry();
+        $entry->setTitle($moduleName);
+        $entry->setDescription($moduleDescription);
+        $entry->setLink($this->url->fromRoute('view-module', $urlParams, ['force_canonical' => true]));
+        $entry->addAuthor(['name' => $module->getOwner()]);
+        $entry->setDateCreated($module->getCreatedAtDateTime());
+        $entry->setId(implode('/', $urlParams));
```
ins0: this looks confused as you don't see in the first place what id is being set here

snapshotpl:
@ins0 what about now? :)

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
```diff
@@ -0,0 +1,80 @@
...
+    public function addModule(ModuleEntity $module)
+    {
+        if (empty($module->getDescription())) {
+            $moduleDescription = 'No description available';
+        } else {
+            $moduleDescription = $module->getDescription();
+        }
+        $moduleName = $module->getName();
+        $urlParams = ['vendor' => $module->getOwner(), 'module' => $moduleName];
+        $id = implode('/', $urlParams);
```
ins0: this is equal odd. how about ``sprintf('%s/%s',  $module->getOwner(), $moduleName)`` or we extend the module entity to get a full repository name - but not sure about that. /cc @Ocramius @localheinz

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @param ModuleEntity $module
+     * @return Entry
+     */
+    public function addModule(ModuleEntity $module)
+    {
+        if (empty($module->getDescription())) {
+            $moduleDescription = 'No description available';
+        } else {
```
ins0: you can skip this ``else`` and declare the first set before the ``if(empty(...``

snapshotpl: Of course I can, but are we really need two assigns to `$moduleDescription`?

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
snapshotpl: I like idea with additional method i module entity. We can reuse it in views.

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
ins0: ocramius suggested this in my pr too, it is cleaner to read - but i'm fine if this stays :smiley:

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
ins0: yeah i ike this idea as well, there should be no reason to follow this way :+1:

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
Ocramius: Yes, two assignments is good

module/ZfModule/src/ZfModule/Mapper/ModuleToFeed.php
Ocramius: If the module pseudo-identifier is this one, then yes, this should be part of the `Module` model API

snapshotpl:
My PR needs review :-)

snapshotpl:
@ins0 who said that Module entity can come from db?

Ocramius:
Merged, thanks @snapshotpl!

PR #416, opened 2015-02-24
Hotfix: Added correct type comparison

### LINKED ISSUE #415: /user Registered Modules bug
Hi!
`/user`in `Registered Module` I see all my repositories, I don't see only registered into site.
It's only a my problem?

Thanks!

ins0:
fixes #415

### DIFF
--- a/module/ZfModule/src/ZfModule/Service/Module.php
+++ b/module/ZfModule/src/ZfModule/Service/Module.php
@@ -131,7 +131,7 @@ class Module extends EventProvider
                 return false;
             }
 
-            if (null === $this->moduleMapper->findByUrl($repository->html_url)) {
+            if (false === $this->moduleMapper->findByUrl($repository->html_url)) {
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


localheinz:
Thank you, @ins0!

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -131,7 +131,7 @@ public function currentUserModules()
                 return false;
             }
 
-            if (null === $this->moduleMapper->findByUrl($repository->html_url)) {
+            if (false === $this->moduleMapper->findByUrl($repository->html_url)) {
```
Ocramius: Urgh, why is a mapper returning `false` on a non-match? :-(

Wondered myself, but apparently that's what it does - I'd personally expect `null`, or a `NullModule`?!

Here's why: https://github.com/zendframework/zf2/blob/8b644e8b0e8cd931bf25cf3babc3ffa047df26d5/library/Zend/Db/ResultSet/HydratingResultSet.php#L93

Ocramius: WTF.

ins0: :train: Doctrine :train:

PR #417, opened 2015-02-24
"Show module on github" target blank

gianarb:
Change target in _blank
Now github does not override zf modules site


localheinz:
Thank you, @gianarb!

gianarb:
:+1:

PR #418, opened 2015-02-24
Fix: Return null if nothing was found by Mapper\Module

localheinz:
This PR

* [x] makes sure that `null` is returned instead of `false` if `ZfModule\Mapper\Module` couldn't find a thing

See https://github.com/zendframework/modules.zendframework.com/pull/416/files#r25241195.

### DIFF
--- a/module/ZfModule/src/ZfModule/Mapper/Module.php
+++ b/module/ZfModule/src/ZfModule/Mapper/Module.php
@@ -146,7 +146,8 @@ class Module extends AbstractDbMapper implements ModuleInterface
         $select = $this->getSelect();
         $select->where([$key => $value]);
 
-        $entity = $this->select($select)->current();
+        $entity = $this->select($select)->current() ?: null;
+
         $this->getEventManager()->trigger('find', $this, ['entity' => $entity]);
 
         return $entity;
--- a/module/ZfModule/src/ZfModule/Service/Module.php
+++ b/module/ZfModule/src/ZfModule/Service/Module.php
@@ -131,7 +131,7 @@ class Module extends EventProvider
                 return false;
             }
 
-            if (false === $this->moduleMapper->findByUrl($repository->html_url)) {
+            if (null === $this->moduleMapper->findByUrl($repository->html_url)) {
                 return false;
             }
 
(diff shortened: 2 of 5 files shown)


localheinz:
@Ocramius 

How about this? Just hang on a second.

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -131,7 +131,7 @@ public function currentUserModules()
                 return false;
             }
 
-            if (false === $this->moduleMapper->findByUrl($repository->html_url)) {
+            if (null === $this->moduleMapper->findByUrl($repository->html_url)) {
```

Instead of a strict comparison here, what do you think, shall we just do this

```php
if (!$this->moduleMapper->findByUrl($repository->html_url)) {}
```

I personally like strict comparisons, but whatever works best should be it.

Ocramius: `!` is indeed enough here: less error prone

module/ZfModule/src/ZfModule/Mapper/Module.php
```diff
@@ -147,6 +147,9 @@ public function findBy($key, $value)
         $select->where([$key => $value]);
 
         $entity = $this->select($select)->current();
+
+        $entity = $entity ?: null;
```
localheinz: @Ocramius 

Does this even make sense, then? Because there are more mappers, and they all return `false`, I bet. Consistency!

Ocramius: That's what `current()` does (in an iterator), heh. A mapper is not an iterator, so I think this ternary is correct

module/ZfModule/src/ZfModule/Mapper/Module.php
```diff
@@ -147,6 +147,9 @@ public function findBy($key, $value)
         $select->where([$key => $value]);
 
         $entity = $this->select($select)->current();
```
Ocramius: Inline the `?: null` here

localheinz:
@Ocramius 

Done!

Ocramius:
:+1:

PR #420, opened 2015-02-24
Fix: Method actually returns unregistered modules

localheinz:
This PR

* [x] renames a method because it actually does something other than what it claims

/cc @gianarb


Ocramius:
:+1:

gianarb:
Thanks! :)

localheinz:
:+1:

PR #429, opened 2015-02-25
Fixed thumbnail size into user/index template

gianarb:
Now my face is responsive!

# Below
below

# After 
after


Ocramius:
:+1:

PR #430, opened 2015-02-25
Basic implementation of centralized bump flow

gianarb:
I was speaking with @localheinz and he explain me the new deploy flow..  

My proposal use [CorleyVersionModule](https://github.com/gianarb/corley-version-module) to centralize a bump of process flow..

* [x] Basic implementation
* [x] update version reference into the application footer
* [ ] write changelog
* [ ] update composer.json `version` key
* ...


config/autoload/global.php
```diff
@@ -1,27 +1,28 @@
...
+            'Zend\\Session\\SessionManager' => 'Zend\\Session\\SessionManager',
+        ),
+    ),
+    'db' => array(
+        'driver' => 'pdo',
+        'dsn' => 'mysql:dbname=modules;host=localhost',
+        'username' => 'modules',
+        'password' => 'modules',
+    ),
+    'version' => '1.1.2',
```
Ocramius: This seems like a bad idea: shouldn't this particular key be in its own config file?

Ocramius: Also: is it actually config? Shouldn't it be just a cached version of the output of something like [`sebastian/version`](https://github.com/sebastianbergmann/version) ?

gianarb: This key is generated and updated by module.. The path of file is configurable

gianarb: This proposal is here for two main ways:

* Show the number of version in production
* Centralized the dump of new application version in one command

Maybe we can change the version key location if the `globa.php` is a bad place :+1:

localheinz:
@gianarb 

This needs a rebase, currently unmergeable!

Also, build has failed.

composer.json
```diff
@@ -13,6 +13,7 @@
         "rwoverdijk/assetmanager": "1.3.*",
         "socalnick/scn-social-auth": "1.14.1",
         "zendframework/zendframework": "~2.3.0",
+        "gianarb/corley-version": "dev-master",
```
localheinz: @gianarb 

Not a bug, but for the overly concerned, there's the `--sort-packages` option which helps keeping required packages in alphabetical order. 

```
$ composer require --sort-packages gianarb/corley-version:dev-master
```

**Disclaimer**: I submitted that feature.

README.md
```diff
@@ -43,3 +43,13 @@ To enable development mode:
 ## Deployment
 
 The master branch of this repository is manually deployed live to [modules.zendframework.com](http://modules.zendframework.com/) by [@GeeH](https://github.com/GeeH).
+
+## Bump new version
+
+```bash
+php public/index.php version-bump <version>
```
localheinz: @gianarb 

Not sure, I think I should have a look at the command, I think I would expect something like:

```
$ php public/index.php version-bump major // bumps the major version
$ php public/index.php version-bump minor // bumps the minor version
$ php public/index.php version-bump patch // bumps the patch version
```

Is this how it works or should work?

config/autoload/global.php
localheinz: @gianarb 

I actually think @Ocramius suggestion about using `sebastian/version` is good - as far as I understand it, the constructor accepts a `$version` and a `$path` to your local repository, and then it will attempt to fetch the version number from the tag, or use `$version` instead.

Regarding the  `version` key in the configuration, I think it is to generic and you probably want to avoid collisions by kind of namespacing your configuration, what do you think?

```php
return [
    'corley_version' => [
        'version' => '1.2.3',
        'path' => __DIR__ . '/../..',
    ],
];
```

Maybe you want to use `sebastian/version` in your helpers?

README.md
gianarb: mm VersioneModule is a simple module to manage bump flow..
In my application for example I use it to manage version of static asset.. In this application this feature in not required.

In this implementation I use it for update a version key into the configuration file and print a version into the footer :) A very stupid feature in this moment.. But we can attach other listeners to implement other "bump steps"
For example we can attach a listener to write changelog of new version :)

Maybe in this moment it's an overhead :)

Ocramius: Let me clarify

 > mm VersioneModule is a simple module to manage bump flow..

There is no such thing as "bumping". There is released software and unreleased software. Bumping as you intend it is not actually useful for the purposes of releasing software.

 > In this implementation I use it for update a version key into the configuration file and print a version into the footer :)

What I actually see is a cachable value that you want to display in the footer without having any stat calls

As I suggested in a chat that we had on hangouts, the best approach would be to have `VersionModule` use `sebastian/version` to read the version and hook into [`Zend\ModuleManager\ModuleEvent::EVENT_MERGE_CONFIG`](https://github.com/zendframework/zf2/blob/b952b35a18d13e3645adcf5721440cbc446d923d/library/Zend/ModuleManager/ModuleEvent.php#L23) and simply write the current version to the cached config.

Additional steps (such as prefixing asset paths and so on) can also be applied there.

The only deployment step to keep in mind is to delete cached files.

Ocramius: That said, I don't think there is a need for an additional deployment CLI command, nor for a new CLI command in general.

gianarb: Yes I remember our discussion :) I start to implement it between a few hours :+1: We can close this PR for now! :+1: I'm working in another branch

PR #433, opened 2015-02-27
Enhancement: Removed asset manager

ins0:
This removes the asset manager. In the current state the asset manager don't provides real effort nor we use multi module public folders.


localheinz:
:+1:

composer.json
```diff
@@ -10,7 +10,6 @@
         "ezyang/htmlpurifier": "4.6.*",
         "monolog/monolog": "~1.12",
         "php": "~5.5",
-        "rwoverdijk/assetmanager": "1.3.*",
```
localheinz: @ins0 

You also need to run

```
$ composer update rwoverdijk/assetmanager
```

ins0: done, some question - what is the point that the ``composer.lock`` file is in the  version control?

public/.gitignore
```diff
@@ -1,4 +1 @@
-img
-css
-js
 favicon.ico
```
localheinz: @ins0 

I think this `.gitignore` can be removed, can't it?

ins0: yeah

composer.json
Ocramius: @ins0 the `composer.lock` MUST be under version control for deployments

public/.gitignore
```diff
@@ -1,4 +0,0 @@
-img
```
Ocramius: Can we actually use bower for this stuff?

localheinz: @Ocramius 

:+1:, but out of scope of this PR, I think!

Ocramius: We are replacing with directly committed assets, heh

Ocramius: Nvm, we are just moving the assets. But yes, we should use bower here.

ins0: :+1:

Ocramius:
:ship: :shipit:

ins0:
thanks @localheinz and @Ocramius

gianarb:
Thanks!!!! @ins0 :+1: :+1:

PR #434, opened 2015-02-27
Enhancement: Added github version to composer 

ins0:
as @gianarb added a version to ``evandotpro/edp-github`` we can add it to composer


Ocramius:
:shipit:

ins0:
:speedboat: thanks @Ocramius

gianarb:
Thanks @ins0! :dancer:

PR #437, opened 2015-03-01
Fix: Update README in regard to releases

localheinz:
This PR

* [x] updates `README.md` with a note that the latest release always represents actual code deployed to production


Ocramius:
:+1:

I believe tagging should be first, deployment should depend on it afterwards, but you have the RM hat here :)

localheinz:
@Ocramius 

Want to update the deployment, too!

PR #442, opened 2015-03-02
Fix: Sudo required for pulling in changes on server

localheinz:
This PR

- [x] updates the `git-pull` task to use `sudo` when actually pulling the changes onto the production server

:bulb: Other than that, the deployment script worked for deploying [1.2.1](https://github.com/zendframework/modules.zendframework.com/releases/tag/1.2.1)!

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -9,7 +9,7 @@
 
     <target hidden="true" name="git-pull">
         <echo message="Pull changes from origin" />
-        <exec logoutput="true" checkreturn="true" command="ssh ${ssh.user}@${ssh.host} -p ${ssh.port} 'cd ${project.root}; git pull origin master ${tag}'" />
+        <exec logoutput="true" checkreturn="true" command="ssh ${ssh.user}@${ssh.host} -p ${ssh.port} 'cd ${project.root}; sudo git pull origin master ${tag}'" />
     </target>
 
     <target hidden="true" name="composer-install">


build.xml
```diff
@@ -9,7 +9,7 @@
 
     <target hidden="true" name="git-pull">
         <echo message="Pull changes from origin" />
-        <exec logoutput="true" checkreturn="true" command="ssh ${ssh.user}@${ssh.host} -p ${ssh.port} 'cd ${project.root}; git pull origin master'" />
+        <exec logoutput="true" checkreturn="true" command="ssh ${ssh.user}@${ssh.host} -p ${ssh.port} 'cd ${project.root}; sudo git pull origin master'" />
```
Ocramius: Seems like a snafu to me. Should at least not operate as root, but eventually as something like the httpd user

localheinz: @Ocramius 

Currently using a user created by @qmanjr5, guess that user should - at best - have these permissions so `sudo` wouldn't be required?

Ocramius: Indeed. We don't allow `sudo` on our machines (for obvious reasons)

localheinz: @Ocramius 

Sent out an :email:!

localheinz:
@Ocramius 

About this one, I was just going to deploy with 

```
$ vendor/bin/phing deploy -Dtag=1.3.0
```

but then the deployment failed on me.

Since we already have `sudo`

* [here](https://github.com/zendframework/modules.zendframework.com/blob/master/build.xml#L17) and
* [there](https://github.com/zendframework/modules.zendframework.com/blob/master/build.xml#L22)

maybe we can just merge it in for now?

Ocramius:
@localheinz we can merge, but those privileges need fixing. `SUDO` is not an option.

Ocramius:
Created #460 for that

PR #446, opened 2015-03-02
Fix: Sort rules in .gitignore alphabetically

localheinz:
This PR

* [x] sorts rules in `.gitignore` alphabetically, both making it more obvious in which way to add new rules, and easier to spot duplicates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,10 @@
 .idea/
 .php_cs.cache
+.vagrant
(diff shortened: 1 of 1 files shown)


localheinz:
@Ocramius 

This is somewhat related to something I'm working on in regard to your https://github.com/zendframework/modules.zendframework.com/pull/433#discussion_r25534414 here!

PR #447, opened 2015-03-02
Contributors

snapshotpl:
Site about us!

PR #451, opened 2015-03-04
Fix: Add missing inline comments and doc-block

localheinz:
This PR

* [x] adds missing inline comments and doc-block to `Controller\ContributorsControllerFactory`

Follows #447.


Ocramius:
:+1:

PR #453, opened 2015-03-04
Enhancement: Display CodeClimate badge in README

localheinz:
This PR

* [x] adds a CodeClimate badge to `README.md`

:bulb: Unfortunately the score is pretty low, but that's mostly because there are a bunch of files (Puppet, but I also guess some of the test with long arrangements) which are *not* excluded from the analysis. In order to set up CodeClimate properly, one apparently needs administrator privileges for a repository. I tried it out the other day for a thing I've been working on and thought it was good.


Ocramius:
Do we also have scrutinizer-ci? It shows a lot of static analysis information that other tools don't really get..

localheinz:
@Ocramius 

I don't mind, we can use Scrutinizer, if you prefer it!

However, I think CodeClimate offers similar insights, you want to take a look?

* https://codeclimate.com/github/localheinz/github-changelog

Ocramius:
@localheinz yeah, this is not a 4.0 codebase, really ;-)

I used CC before and it's simple and nice, but really doesn't show what is relevant to us.

localheinz:
@Ocramius 

Hang on, I'm checking out Scrutinizer!

localheinz:
@Ocramius 

screen shot 2015-03-04 at 18 02 39

:cry:

Ocramius:
Meh, always requires repo settings :-(

PR #455, opened 2015-03-04
Enhancement: Implement Entity\Repository

localheinz:
This PR

* [x] implements `Application\Entity\Repository`
* [x] implements a `Service\GitHubRepositoryFactory` returning an `Entity\Repository` with values retrieved from configuration
* [x] adjusts `Controller\ContributorsController` and `View\Helper\GitHubRepository` to rely on that entity, rather than a configuration array

Follows https://github.com/zendframework/modules.zendframework.com/pull/447#discussion_r25689320.

### DIFF
--- a/module/Application/config/module.config.php
+++ b/module/Application/config/module.config.php
@@ -97,6 +97,7 @@ return [
     'service_manager' => [
         'factories' => [
             \HTMLPurifier::class => Service\HtmlPurifierFactory::class,
+            'github_repository' => Service\GitHubRepositoryFactory::class,
             Log\LoggerInterface::class => Service\LoggerFactory::class,
             Service\ErrorHandlingService::class => Service\ErrorHandlingServiceFactory::class,
             Service\RepositoryRetriever::class => Service\RepositoryRetrieverFactory::class,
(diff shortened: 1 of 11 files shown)


module/Application/config/module.config.php
```diff
@@ -99,6 +99,7 @@
     'service_manager' => [
         'factories' => [
             \HTMLPurifier::class => Service\HtmlPurifierFactory::class,
+            'github_repository' => Service\GitHubRepositoryFactory::class,
```
Ocramius: Why the different naming?

Ocramius: Naming should probably be `project_github_repository` or such

localheinz: @Ocramius 

Not sure, I guess because it's mostly used by `View\Helper\GitHubRepositoryUrl`?
Do you have a suggestion? I'm happy to hear it!

localheinz: @Ocramius 

Hang on a second!

localheinz:
@Ocramius 

Implemented the changes as suggested!

Ocramius:
:+1:

localheinz:
@Ocramius 

Awesome, thank you for reviewing - I'm now going to have a look at @gianarb's PRs!

PR #462, opened 2015-03-05
[WIP] Fix: Regular expression in route definition

localheinz:
This PR

* [x] adds a failing test, asserting the route doesn't match if the user name is invalid
* [x] fixes a config key which should be `constraints` instead of `constrains`
* [ ] fixes the regular expression for the route match with the feedback of @Ocramius and @ins0 

Follows #444.

PR #444, opened 2015-03-02
Feature/modules from user

gianarb:
Reference to #431

PR #464, opened 2015-03-05
Fix: Integration test for User controller fails because of lack of mocking

localheinz:
This PR

* [x] adds a failing assertion
* [x] mocks `Mapper\User` to fix the failing test
* [x] introduces a variable to avoid duplication

Follows #444.
Spotted in #462.


localheinz:
@Ocramius 

Need to go now, will work on this later!

PR #466, opened 2015-03-05
Fix: Config key should be 'constraints' instead of 'constrains'

localheinz:
This PR

* [x] adds a failing test
* [x] fixes a config key, which should be `constraints` instead of `constrains`

Related to #462.


Ocramius:
:+1:

PR #467, opened 2015-03-05
Fix: Remove useless route definition

localheinz:
This PR

* [x] removes a useless route definiton

:bulb: This is probably a remainder of the Zend Skeleton application, isn't it?


Ocramius:
:+1:

PR #468, opened 2015-03-05
Fix: organizationAction() should be listAction()

localheinz:
This PR

* [x] renames `Controller\IndexController::organizationAction()` to `listAction()` because that's what is in the route name, and also, what the action does, that is, list modules

:bulb: Controller should probably be named differently as well, e.g., `ModuleController`


Ocramius:
Requires rebasing

localheinz:
@Ocramius 

Rebased, thank you for reviewing!

Ocramius:
:+1:

localheinz:
Thank you, @Ocramius!

PR #470, opened 2015-03-05
Fix: Rename IndexController to ModuleController

localheinz:
This PR

* [x] renames `IndexController` to `ModuleController` as this is a more specific name for what it does, that is, controlling listing, viewing, adding and removing modules
* [x] renames folder containing the corresponding view scripts

Related to #468.


Ocramius:
:+1:

PR #421, opened 2015-02-24
[WIP] Feature: Flash Messenger Error Messages

ins0:
This PR

- [x] Replaces the generic error exceptions with specific exception constructors
- [x] Extend the exceptions with a public message
- [x] Covers all flashmessenger messages in a clean bs message format
- [x] Covers the error view in a clean bs message format + added public messages from exceptions
- [x] Covers the 404 error in a clean bs message format

PR #475, opened 2015-03-06
Enhancement: added hhvm as allow failure

ins0:
as the page runs not on ``hhvm`` it should be marked as ``allow_failures``. 

this would also be required as of pr #421 and the ``hhvm`` exception bug in phpunit - as an fix is not landed in phpunit yet.

PHPUnit Bug: https://github.com/sebastianbergmann/phpunit-mock-objects/issues/207

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ php:
   - 5.6
   - hhvm
 
+matrix:
+  fast_finish: true
+  allow_failures:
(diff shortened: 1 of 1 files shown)


localheinz:
@Ocramius 

Do you have any objections? I'm fine with it!

localheinz:
Thanks, @ins0!

PR #476, opened 2015-03-06
Fix: Do not collect code coverage . . . for now

localheinz:
This PR

* [x] removes configuration which results in code coverage being collected when running tests

:bulb: We're not using the code coverage data generated on Travis *yet*, so it's a waste of resources (and time). Removing the configuration significantly speeds up the tests, and if one is interested in code coverage, you can still specify a command line option or run the tests from within your favourite IDE™.

#### Before

screen shot 2015-03-06 at 14 31 32

#### After

screen shot 2015-03-06 at 14 31 45

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 .php_cs.cache
 .vagrant
 build.properties
-clover.xml
 config/development.config.php
 data/logs/*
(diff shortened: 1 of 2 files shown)


gianarb:
:100:

PR #484, opened 2015-04-25
Updated contribute to reflect new gitflow process

GeeH:


localheinz:
@GeeH 

Not sure why we would need a `develop` branch, can you explain? We only removed it a few months ago in order to simplify the process.

GeeH:
It was removed because there was a a shambles in the repo, I’d like to use traditional gitflow process so we can have releases cut from master and develop having the features ready for next release. I’d like to formalise the release process now we have stability because some of the new features need database changes before they can be released (see my latest PR).

Saying that, I’m happy to just use master, but we’d need to co-ordinate PR and deployment more closely.

g 


On 25 April 2015 at 15:28:38, Andreas Möller (notifications@github.com) wrote:

@GeeH

Not sure why we would need a develop branch, can you explain? We only removed it a few months ago in order to simplify the process.

GeeH:
Any thoughts @localheinz? As I said, I'm happy to trash this idea if we can get some more co-ordination on releases.

localheinz:
@GeeH 

Here's how the current process works:

* every branch gets merged directly into `master`
* a deployment is represented by a tag, meaning that you create a tag first, then deploy - thus, the latest tag always represents the code currently deployed to production

What do you think?

Ocramius:
What @localheinz suggests is simpler - I wouldn't maintain gitflow on
public repos without too many contributors unless there is a clear benefit
(too many things happening at once)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

gianarb:
In this moment in my opinion git flow is an overhead..
I hope that number of contributors grow :)

GeeH:
Agreed - closed.

PR #485, opened 2015-04-25
Removed Twitter box

GeeH:
I don't think the Twitter box is appropriate - we are still piggy backing on Zend's branding, without checking that Zend (or possibly the CR team) are OK with it. I also think that the Twitter account is currently giving no value as nobody is really tweeting with it, and it's in control of one person (which is very dangerous). I'm happy to re-instate this in the future if we get sign off from Zend.

### DIFF
--- a/module/Application/view/application/index/index.phtml
+++ b/module/Application/view/application/index/index.phtml
@@ -55,15 +55,6 @@
                 </div>
             </div>
         </div>
-        <div class="block">
-            <h4><strong>Twitter</strong></h4>
-            <hr>
-            <div class="row">
-                <div class="col-md-12">
-                    <?php echo $this->twitterWidget(); ?>
-                </div>
-            </div>
-        </div>
         <?php echo $this->newUsers(); ?>
     </div>
 </div>


gianarb:
@GeeH what is a good way to share twitter's account? :) 
If you send me an email I share credential and other stuff.. (All account if this stuff is important for you..)

I can remove ZF logo from avatar if it's a problem..
This account is only a sharing opportunities for our modules
ping @weierophinney for this stuff :) Thanks

GeeH:
I understand you are keen to do this, but we've can just assume it's acceptable to Tweet under Zend's brand. We need to get permission FIRST and then do these things.

gianarb:
ohh yes :) I use ZF logo only with avatar, we can change it :)

GeeH:
You don’t understand, people will assume that it’s an official Zend account as we are using ZF - short for Zend Framework - in the account name. We need to get that approved by Zend before we can do that.


On 25 April 2015 at 10:56:49, Gianluca Arbezzano (notifications@github.com) wrote:

ohh yes :) I use ZF logo only with avatar, we can change it :)

gianarb:
Ok :) No problem..

PR #486, opened 2015-04-25
Removed vagrant and puphpet scripts

### LINKED ISSUE #441: Remove puphpet?
What are you thing about remove puphpet files? Currently it's 12 MB of configuration files and scripts. We can still support vagrant using puphpet. How? Just keep `puphpet/confg.yaml` and `puphpet/files/*/*.sh`, other files should be ignored and can be generated manually by https://puphpet.com/ site.

GeeH:
Closes #441 

I don't think the current vagrant and puphpet process works or adds any value, so I propose removing it and doing it again if it's needed.


gianarb:
:+1: go go go!

adamlundrigan:
I've become so reliant on vagrant for dev environments I don't know how to go on living without it :P

How about something based on Vagrant and docker?  I've used [a setup similar to this one](https://github.com/kasperisager/phpstack) on a few projects and it works well.  It's very lightweight in terms of what you have to store in the repository ([Vagrantfile](https://github.com/kasperisager/phpstack/blob/master/Vagrantfile), [Dockerfile](https://github.com/kasperisager/phpstack/blob/master/images/nginx/Dockerfile), [docker-compose.yml](https://github.com/kasperisager/phpstack/blob/master/docker-compose.yml)).  This site doesn't need much more than a standard LAMP stack so the whole thing should come in around 200 lines across the three files.

Ocramius:
The problem is not vagrant tbh, but just @puphpet, which really is unmaintainable once dragged into a project :-(

adamlundrigan:
Yup, It's a pain and makes all my repository language stats look like this:

image

I'll put together something and send in a PR

PR #487, opened 2015-04-25
Feature/add scrutinizer

### LINKED ISSUE #459: Consider using Scrutinizer and displaying corresponding badges
I think @Ocramius and I agree that is a good idea to use Scrutinizer (or a similar tool) for performing code quality / coverage checks in addition to what we already have, and then also display the corresponding badges in `README.md`.

However, setting this up requires **ADMIN** privileges, so if anyone with those would be willing to set it up, that would be awesome!

See https://github.com/zendframework/modules.zendframework.com/pull/453#issuecomment-77198577.

/cc @EvanDotPro @ezimuel  @weierophinney 

### Scrutinizer

screen shot 2015-03-04 at 18 02 39

GeeH:
Closes #459


GeeH:
Note, the config files may need tweaking - but I think this is a good start.

PR #489, opened 2015-04-27
Removed Twitter box

GeeH:


localheinz:
@GeeH 

Do you want to remove the corresponding view helper and view scripts as well?

* [`ZfModule\View\Helper\TwitterWidget`](https://github.com/zendframework/modules.zendframework.com/blob/master/module/ZfModule/src/ZfModule/View/Helper/TwitterWidget.php)
* [`zf-module/helper/twitter-widget-view.phtml`](https://github.com/zendframework/modules.zendframework.com/blob/master/module/ZfModule/view/zf-module/helper/twitter-widget-view.phtml)

?

GeeH:
No, we may add this back but we need to put the Twitter account under some kind of rules

localheinz:
@GeeH 

Ok, then :rocket:!

PR #490, opened 2015-04-27
Removed vagrant and puphpet scripts

GeeH:


localheinz:
@GeeH 

Looks good to me, so I'm merging this in!

PR #492, opened 2015-04-27
Feature/add module rating - part 1 of 2

GeeH:


bin/cli.php
```diff
@@ -0,0 +1,36 @@
...
+chdir(__DIR__ . '/../');
+require 'vendor/autoload.php'; // grabs the Composer autoloader
+error_reporting(E_ALL ^ E_WARNING);
+set_time_limit(0);
+
+$application = Zend\Mvc\Application::init(require 'config/application.config.php');
+
+$services = $application->getServiceManager();
+
+$updateModel = $services->get(Update::class);
```
localheinz: @GeeH 

Unsure about the naming here - I understand `ZfModule\Service\Update` as a service, not a model?

GeeH: It's not a model, it's a service.

module/ZfModule/src/ZfModule/Service/Update.php
```diff
@@ -0,0 +1,131 @@
...
+        $module->setUpdatedAt(date('Y-m-d H:i:s'));
+
+        $this->moduleMapper->update($module);
+    }
+
+    /**
+     * @param ModuleEntity $module
+     * @return array|bool
+     */
+    public function getFromGithub(ModuleEntity $module)
```
ins0: wouldn't it be better to handle this stuff in the ``RepositoryRetreiver`` ?

module/ZfModule/src/ZfModule/Service/Update.php
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        if (!is_array($data)) {
+            return false;
+        }
+
+        return $data;
+    }
+
+    private function setupClient()
```
ins0: see above

GeeH: Nah - EdpGithub has a dependency on ScnSocialAuth which has a dependency on some omni-auth component. Long and the short is it only works with Http requests and I cannot be bothered to fix that.

localheinz:
@GeeH 

Not sure, what is the state of this, shall we close it, or would you prefer to continue working on this?

GeeH:
@localheinz 

No idea, if I'm honest, I think this whole project is a shitshow that probably needs to be either rewritten from scratch, or shut down.

Ocramius:
Agree here: serves no purpose on its own, even more now that packagist supports most of what we'd look for in it.

localheinz:
@GeeH @Ocramius 

Not sure, should we shut the site down, then?

Ocramius:
I think so. Maybe freeze it into static mode, to preserve links.

Issue #496, opened 2015-09-14
Changing the current DB structure


ins0:
I would like to suggest and create a PR with a new DB structure.

The current design isn't nice at all and cause trouble along the way. The modules table isn't well normalized and cause trouble when it comes to "which user uploaded a module" - user repo? company repo? or when a user is changing his github name (not possible currently). On the other side i would like to prepare the site to support different modules from ZF2 and upcoming ZF3.

I will upload the suggested db image today when i'm back home.

So in the meantime i would like to get some feedback if you think this is a good or bad idea and if we should switch with this change to a ORM (doctrine).

/cc @Ocramius @localheinz @GeeH @adamlundrigan @gianarb

gianarb:
Hi @ins0 
This proposal sounds good.. This is a strong feature :)

If we would implement doctrine we can use all other tools (migrations, fixtures) to increase our workflow

ins0:
True @gianarb - i think we already tried to implement migrations but i'm not quite sure. I added a first suggestion for the db.

localheinz:
@gianarb @ins0 

I'd second the use of Doctrine, and I'm happy to help!

localheinz:
@gianarb @ins0 

Here's probably how I would do this, as it's a bit of work:

1. switch to using `doctrine/orm` in the first place
2. start using `doctrine/migrations`
3. make changes to the schema

What do you think? All at once, that's too much - and there's probably hardly anyone willing to code review that. Let's take small steps!

gianarb:
@ins0 now we have migrations :P Can we apply this change in your opinion?

PR #497, opened 2015-09-19
[WIP] Feature/doctrine

ins0:
This PR is adding Doctrine ORM to the Projekt and will replace the Zend\Db Mapper.
Follows after Discussion in #496 

- [x] added doctrine
- [x] added module entity
- [ ] added user entity
- [x] changed application module
- [ ] changed user module
- [ ] changed zfmodule module
- [ ] remove old mapper
- [ ] edit existing tests

@localheinz @gianarb


localheinz:
@ins0 

What do you think, can I help with a PR that prepares for this and adds the capability to run Doctrine migrations?

module/Application/src/Application/Controller/SearchController.php
```diff
@@ -4,28 +4,28 @@
...
-        $this->moduleMapper = $moduleMapper;
+        $this->moduleService = $moduleService;
     }
 
     public function indexAction()
     {
         $query =  $this->params()->fromQuery('query', null);
 
-        $results = $this->moduleMapper->findByLike($query);
+        $results = $this->moduleService->findModules($query, ['m.name']);
```
ins0: @localheinz not sure about this but is it legal to place this param with his specific prefix ``m.name`` in the controller?

module/ZfModule/src/ZfModule/Service/Module.php
```diff
@@ -24,16 +27,62 @@ class Module extends EventProvider
...
+    /**
+     * Get all Modules
+     *
+     * @param null $limit
+     * @param null $offset
+     * @return array
+     */
+    public function getModules($limit = null, $offset = null)
+    {
+        return $this->moduleRepository->findAll($limit, $offset)->getResult();
```
ins0: @localheinz i prefer to return a query from repositorys rather than the direct results - good? bad?

ins0:
> What do you think, can I help with a PR that prepares for this and adds the capability to run Doctrine migrations?

Yes please! I'm not very familiar with Doctrine Migrations nor how to setup this feature

module/Application/src/Application/Controller/SearchController.php
localheinz: @ins0 

It's probably better not to expose internals to much - so how about (if need be) exposing a method which fill find modules by name, specifically (if that's what it does)?

module/ZfModule/src/ZfModule/Service/Module.php
localheinz: @ins0 

Here's blog post which might help: http://www.whitewashing.de/2013/03/04/doctrine_repositories.html!

Ocramius: @ins0 really bad idea to return queries from repositories

ins0: thanks @Ocramius @localheinz i will change this. One question, DQL or Querybuilder?

module/Application/src/Application/Controller/SearchController.php
ins0: that's what i thought too, i will change this, ty

localheinz:
@ins0 

Doctrine migrations are implemented in #502.

About this PR, what do you think about breaking it up into smaller parts? Since we agreed that we want to use Doctrine, it doesn't need to be implemented all in one PR.

Maybe start with creating entities and making sure we've got the schema right? Then, in a next step, start using them.

We should be running

```
$ php public/index.php orm:info
$ php public/index.php orm:validate-schema
```

as part of the `script` on Travis, and just make sure for a start that the mappings are right.

What do you think?

ins0:
@localheinz sound good to me. what do u suggest, changing the current db structure - creating doctrine entitys or vice versa?

This is the first time i meet doctrine migrations (oh boy - don't let @Ocramius hear that :smile_cat: ) so  currently i don't know for sure what had changed, if someone can give me a short hint with this that would be nice.

/cc @gianarb

gianarb:
IMO we can use doctrine\migrations to change the current database schema.
After this process we can write a correct entities.. :)

A migration is very easy to implement.
[http://docs.doctrine-project.org/projects/doctrine-migrations/en/latest/reference/migration_classes.html](docs).

We can write a SQL script to update our struct.. And another SQL script to down it in the old version

localheinz:
@ins0 

Depends entirely on which way you want to go - do you have a preference?

By the way, [`zendframework/modules.zendframework.com:1.6.0`](https://github.com/zendframework/modules.zendframework.com/releases/tag/1.6.0) has been deployed - with a little downtime, as I forgot to edit the local configuration files *before* deploying, but, it's up!

localheinz:
@ins0 

Following the merge of #502, you need to rebase, sorry!

localheinz:
@ins0 

In regard to the migrations, as @gianarb suggested, you can start by creating migrations as suggested:

```php
$ php public/index.php migrations:generate
```

This will create a migration class, which you can modify to your needs. 

If you have a migration which you can't or don't ever want to be able to roll back, you should probably throw a [`Doctrine\DBAL\Migrations\IrreversibleMigrationException`](https://github.com/doctrine/migrations/blob/v1.1.0/lib/Doctrine/DBAL/Migrations/IrreversibleMigrationException.php).

ins0:
thanks @gianarb @localheinz , so i think i will stop here and handle the remaining todos (described above) in different prs as this is more clear, do u agree?

gianarb:
Yes.. It sounds good :)

More easy to merge and to follow :)

2015-10-19 11:11 GMT+02:00 Marco Rieger <notifications@github.com>:

> thanks @gianarb <https://github.com/gianarb> @localheinz
> <https://github.com/localheinz> , so i think i will stop here and handle
> the remaining todos (described above) in different prs as this is more
> clear, do u agree?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/modules.zendframework.com/pull/497#issuecomment-149153460>
> .
>



-- 
Gianluca Arbezzano
www.gianarb.it

PR #502, opened 2015-09-26
Enhancement: Use doctrine/migrations to modify the database schema

localheinz:
This PR

* [x] requires `doctrine/doctrine-module`
* [x] requires `doctrine/doctrine-orm-module`
* [x] requires `doctrine/migrations`
* [x] adds a basic configuration for the `DoctrineORMModule` (minimum required for migrations)
* [x] enables both `DoctrineModule` and `DoctrineORMModule`
* [x] extracts a migration from `data/sql/0.sql`
* [x] uses migrations to set up database in tests
* [x] removes the database dump
* [x] adds a test case for migrations
* [x] moves migrations into `Application` namespace
* [x] updates build target to run migrations in production, too

Related to #497.

:information_desk_person: The Doctrine console commands are available via

```
$ php public/index.php
```

Enjoy!

PR #503, opened 2015-09-26
Fix: Make sure to fix *.php.dist, too

localheinz:
This PR

* [x] adds `*.php.dist` to the patterns of file names which should be fixes
* [x] runs `make cs`

Related to #502.


gianarb:
can we merge this pr? :)

Ocramius:
:+1:

PR #504, opened 2015-10-17
Fixed forgot chapter

gianarb:
Vagrant at the moment is not supported


Ocramius:
Huh? Since when? :|

gianarb:
We removed phpuppet a lot of time ago :)

Ocramius:
Yeah, and there's no Vagrant support at all now? Like, not even with simple bash provisioning? Can we at least fix that? Of course people won't be able to contribute otherwise :|

gianarb:
Ok I build a slim solution.

Close #504

2015-10-18 21:01 GMT+02:00 Marco Pivetta <notifications@github.com>:

> Yeah, and there's no Vagrant support at all now? Like, not even with
> simple bash provisioning? Can we at least fix that? Of course people won't
> be able to contribute otherwise :|
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zendframework/modules.zendframework.com/pull/504#issuecomment-149039203>
> .
>



-- 
Gianluca Arbezzano
www.gianarb.it

Issue #505, opened 2015-10-19
I hate config autoload directory


gianarb:
Hello guys..
This night I dreamed this directory.

![screen shot 2015-10-19 at 21 47 55](https://cloud.githubusercontent.com/assets/1630267/10589339/1a360fe2-76ab-11e5-91ae-315a056d745c.png)

In your opinion this is a good practice?
Can we try to replace decrease the number of files?

localheinz:
@gianarb 

Here's how I would do it:

* one `global.php` that contains all of the configuration
* if you need to override configuration, create a `local.php`, override it

this many files is kind of ugly.

I understand, though, that `development.local.php.dist` is used by `ZF\DevelopmentMode`, though, so this one needs to stay, apparently.

gianarb:
:+1: I agree with you

PR #506, opened 2015-10-19
Fix: Merge global configuration files

localheinz:
This PR

* [x] merges global configuration files
* [x] keeps main keys sorted alphabetically
* [x] makes use of the class keyword

Related to #505.


gianarb:
IMO now is better! Thanks

localheinz:
Thank you, @gianarb!

PR #510, opened 2015-10-27
Enhancement/Removed hhvm runs

ins0:
I think we all agree that the Travis HHVM runs don't accomplish anything and only use resources that we really don't need. 

This PR

- [x] removes Travis HHVM runs
- [x] adds PHP7 Travis runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ sudo: false
 
 php:
   - 5.6
-  - hhvm
+  - 7
 
 matrix:
   fast_finish: true
   allow_failures:
-    - php: hhvm
+    - php: 7
 
 cache:
   directories:


Ocramius:
@ins0 why did you remove 7?

ins0:
@Ocramius the php7 runs allways fails, as of the required composer dependencies. Should it stay?

Ocramius:
I think it should, as we will move to it when possible :|

ins0:
@Ocramius ok i was on the don't waste Travis Resources Plan, added 7 support :)

Ocramius:
@ins0 thanks!

ins0:
thanks @Ocramius

PR #511, opened 2015-10-30
Build Vagrant slim configuration

gianarb:
This commit adds new vagrantfile without puppet, it use shell
provisioning to build a develop box with mysql and apache


ins0:
nice @gianarb could you please update the readme file as well?

gianarb:
Can we test it?! Thanks

PR #514, opened 2017-05-17
Rename .php_cs to .php_cs.dist

keradus:


.travis.yml
```diff
@@ -28,7 +28,7 @@ before_script:
   - php public/index.php migrations:migrate --no-interaction --verbose
 
 script:
-  - vendor/bin/php-cs-fixer fix --config-file=./.php_cs --dry-run --diff -v
+  - vendor/bin/php-cs-fixer fix --config-file=./.php_cs.dist --dry-run --diff -v
```
keradus: in PHP CS Fixer config, it already asking to fixing `->name('*.php.dist')`, while `.dist` suffix is not in filename on master branch :D

